Issue #11, opened 2026-05-18
Add server-side WordPress metrics — plugin count, hook count, DB sizes, order table health


Ajay-Singh-Adhikari:
## Problem

The plugin currently only collects WooCommerce order metrics and autoloaded options. When an engineer installs the plugin on a new client site, the dashboard shows order table health — but nothing about why the site might be slow beyond the database.

Three of the most common WordPress performance problems are completely invisible today: too many plugins loading on every request, too many hooks firing per page, and unexpected growth in non-WooCommerce database tables. An engineer has to SSH in, run WP-CLI commands, and query the database by hand to find any of this. Every new client engagement starts from zero.

---

### Plugin count (active + inactive)

**Why monitor it:**
Every active plugin loads PHP code on every request. A site with 50+ active plugins uses more memory, has more potential conflicts, and fires more code on every page load. Plugin count is the fastest signal for extension footprint — "you have 54 active plugins" immediately explains slowness without needing to dig further.

**What you can decide after collecting it:**
- Whether the performance problem is likely code-layer (too many plugins) or data-layer (DB bloat)
- Whether a plugin audit is needed before any other optimisation
- If plugin count increases over time in BigQuery, someone is adding plugins without oversight — catch it early
- Future: trigger an alert when plugin count crosses a configured threshold

### Hook count per page

**Why monitor it:**
WordPress fires actions and filters on every page load. Every plugin registers hooks. A site with 2000+ hooks per page is almost always slow — hook count correlates directly with page generation time. Right now the plugin sees DB symptoms but has no visibility into the application layer at all.

**What you can decide after collecting it:**
- Whether the problem is application-layer (too many hooks) or data-layer (DB size) — this distinction determines the entire remediation path
- If DB metrics are healthy but QET is high and hook count is also high, the fix is code review not archival
- If hook count grows over time, a recently added plugin is the likely culprit
- Helps prioritise: do we optimise the DB first or audit the plugin stack first?

### DB table sizes (all tables via information_schema)

**Why monitor it:**
The plugin currently only tracks WooCommerce tables. A client could have a 40GB database where only 5GB is order data — a logging plugin, a badly built custom table, or years of post revision bloat is consuming the rest. Without a full table breakdown, the plugin is blind to everything outside WooCommerce. An engineer could archive all orders and still have a slow database because an entirely different table was the real problem.

**What you can decide after collecting it:**
[...]

PR #13, opened 2026-05-19
feat: add WooCommerce order table metrics

### LINKED ISSUE #12: Collect order table size, oldest order age and show a WooCommerce table in Wp Admin Pro-Perf Dashboard
## Problem

WooCommerce order tables grow with every transaction and are never automatically cleaned up. Over time, `wp_woocommerce_order_item` and `wp_woocommerce_order_itemmeta` can become the largest tables in the database and a significant contributor to slow query execution.

Today there is no automated signal for this. An engineer has to manually query the database to find out how large these tables are and how far back the order history goes. Every engagement discovers this from scratch.

## What we are missing

| Signal | What it tells you | How to collect |
|---|---|---|
| Order table size | How much storage order data is consuming | `information_schema` query on `wp_woocommerce_order_item` and `wp_woocommerce_order_itemmeta` |
| Oldest order date | How far back the data goes relative to the client&#39;s retention window | `MIN(date_created)` on the order table |

## Why this matters

Large order tables with years of unarchived data directly impact query performance. Archiving old data is a proven way to reduce table size and improve backend execution — but the decision to archive requires data first: how big are the tables, and how old does the data go back?

## Why table size alone is not a reliable trigger

A low-transaction-volume site can accumulate over a decade of order history while keeping table sizes under 20 MB. A &#34;size &gt; limit&#34; trigger would never fire for that site even though orders are 10+ years old.

This makes table size alone an unreliable archival trigger across clients with different transaction volumes.

## Agreed direction for the archival trigger

Based on the above, the adapter will use **query execution time (QET)** as the primary performance signal, combined with an age check:

> Archive when: QET &gt; 2× post-archival baseline **AND** oldest order is older than the configured retention window

Two plugin settings back this trigger:
1. **Order retention window (years)** — default: 2. The client sets this based on how much order history they need to show customers.
2. **Table size limit (MB)** — optional upper bound surfaced as a context metric, not used as the primary trigger condition.

The archival trigger implementation is tracked in #14. The QET collector is tracked in #15.

## Expected outcome

The adapter collects and pushes to BigQuery on every scheduled run:

**Already collecting:**
- Autoload option count
- Autoload option size

**Added by this issue:**
- Size of `wp_woocommerce_order_item` (MB)
- Size of `wp_woocommerce_order_itemmeta` (MB)
- Date of the oldest order in the database

## What changes after this

- An engineer opens the WP Admin ProPerf dashboard and immediately sees how large the order tables are and how far back the data goes — no manual DB query needed.
- The archival conversation with the client starts from data, not guesswork.
- A WooCommerce Order Metrics table is visible in the WP Admin ProPerf dashboard showing order items table size, order itemmeta table size, oldest order date, and oldest order age.

## Related
- Parent tracking issue: [performance-adapter-wp #11](https://github.com/ColoredCow/performance-adapter-wp/issues/11)
- Platform-agnostic layer: [performance-dashboard #1](https://github.com/ColoredCow/performance-dashboard/issues/1)
- Archival trigger: #14
- QET collector: #15

Ajay-Singh-Adhikari:
## Target

Closes #12 — WooCommerce order table metrics (archival signal)

## What changes this PR has

- Added `woo_order_items_size_mb` — shows how much storage the order items table is consuming. Using this we can determine if the table size is large enough to affect query execution time.
- Added `woo_order_itemmeta_size_mb` — same signal for the order itemmeta table, which is typically the larger of the two.
- Added `woo_oldest_order_date` — shows how far back the order history goes, so we can decide if old data is a candidate for archival.
- Returns `0.0` for table sizes and `null` for oldest order date when the WooCommerce tables are not present (e.g. WooCommerce not active on the site).

## Expected outcome

- A new WooCommerce Order Metrics table is visible in the WP Admin section alongside the existing Autoloaded Options table.
- The table shows 5 rows: order items table size, order itemmeta table size, oldest order date, oldest order age, and archival trigger.
- Clicking "Push to BigQuery" from WP Admin sends the real metrics to BigQuery — all three new `woo_*` fields appear in the BigQuery table alongside the existing autoload metrics.


properf-wordpress-adapter.php
```diff
@@ -222,6 +222,16 @@ function properf_render_dashboard() {
...
 	$top_size_keys = $autoloaded_data_metrics['top_size_keys'];
 
+	$woo_metrics        = $metrics['woo'];
+	$oldest_date        = $woo_metrics['oldest_order_date'];
+	$oldest_age_days    = null;
+	$archival_triggered = false;
+
+	if ( $oldest_date ) {
+		$oldest_age_days    = (int) floor( ( time() - strtotime( $oldest_date ) ) / DAY_IN_SECONDS );
+		$archival_triggered = $oldest_age_days >= 365;
```
pokhiii: Saw our chat thread about this earlier — thanks for thinking through the days-based vs date-based tradeoff carefully. The AND-condition design you proposed (`table size > limit AND oldest order older than N days`) makes a lot of sense, and the SankalpTaru data point — 13 years of orders, only ~19 MB — is a really good reminder that table size alone isn't a strong enough signal on its own. Nice catch.

One thing I want to flag though — the description of issue #12 still says:

> "The adapter does not enforce any rule. It surfaces the raw numbers so the engineer and client can decide together. The archival window is a business decision made with each client."

But our chat moved past that. We are now planning a configurable trigger with two conditions, two new plugin settings (retention days + table size limit), default values, and maybe a query-execution-time collector after that. That is a real scope change from what the issue documents.

Before this PR lands, could we make issue #12 the source of truth?

1. Could you update the issue #12 description to reflect the new direction — the two plugin settings, the AND-condition trigger, the default values, and the reasoning (the SankalpTaru observation is great context to include)?
2. The query-execution-time collector you described at the end of our chat sounds valuable, but it is a different signal from "order table archival". Could that get its own issue under #11? Easier to scope and review on its own, and keeps each PR focused on one thing.

A general rule I'd suggest for us going forward: if we agree on something in chat that changes the requirement, we update the GitHub issue right after. Otherwise the issue and the code drift apart over time, and a reviewer who joins later (or future-you reading the issue 6 months from now) won't know what was actually agreed.

For this PR specifically — I'd suggest keeping the scope small: just the three raw signals you already have (two table sizes + oldest order date), and ship the trigger + plugin settings in a follow-up PR after #12 is updated. The hardcoded `>= 365` here doesn't match the new design either (no plugin setting, no size check), so taking it out of this PR makes sense regardless.

What do you think?

pokhiii (commented):
Nice work on this — the new collector follows the existing autoload pattern closely, the size queries use `$wpdb->prepare()` correctly, and the dashboard table reads cleanly. Left a few inline questions / thoughts, mostly around two things:

1. The hardcoded 365-day rule and the "Archival Trigger" row — I think the issue wants the adapter to stay rule-free and just show raw numbers.
2. WooCommerce HPOS (High-Performance Order Storage) — on modern WC sites the oldest-order query may not find any data because orders live in a different table now.

None of these are blocking once we talk them through. Curious to hear your take on the inline threads.

includes/class-data-collector.php
```diff
@@ -67,6 +70,51 @@ public function collect_autoloaded_options() {
...
+				FROM information_schema.TABLES
+				WHERE table_schema = %s
+				AND table_name = %s',
+				$db_name,
+				$wpdb->prefix . 'woocommerce_order_itemmeta'
+			)
+		);
+
+		$oldest_order_date = $wpdb->get_var(
+			"SELECT MIN(post_date_gmt) FROM {$wpdb->posts} WHERE post_type = 'shop_order' AND post_status != 'trash'"
```
pokhiii: Quick question — have you tried this on a WooCommerce site that has **HPOS** turned on? 

A short bit of context in case it's new: HPOS stands for *High-Performance Order Storage*. It's a newer way WooCommerce stores orders, and it's the default since WooCommerce 8.x. On HPOS sites, orders are **not** stored in `wp_posts` with `post_type = 'shop_order'` anymore — they live in a separate table called `wp_wc_orders`, with a column `date_created_gmt`.

So this query would return `NULL` on any HPOS site, and the `oldest_order_date` signal would silently be missing for them — which is the most useful signal for the archival conversation that issue #12 is about.

Reading material if you want to dig in:
- https://developer.woocommerce.com/docs/2022/09/12/high-performance-order-storage/
- https://github.com/woocommerce/woocommerce/wiki/High-Performance-Order-Storage-Upgrade-Recipe-Book

One way to handle both cases:

```php
if ( class_exists( '\Automattic\WooCommerce\Utilities\OrderUtil' )
     && \Automattic\WooCommerce\Utilities\OrderUtil::custom_orders_table_usage_is_enabled() ) {
    // HPOS: query {$wpdb->prefix}wc_orders.date_created_gmt
} else {
    // Legacy (CPT): the existing wp_posts query
}
```

Good news: the two table-size signals above (`woocommerce_order_items` and `woocommerce_order_itemmeta`) don't need to change — those tables still exist under HPOS because they hold the **line items** per order (the individual products inside an order), not the order headers themselves. So only this third query needs the HPOS branch.

Also a small detail — issue #12 actually says `MIN(date_created)` on the order table, which already sidesteps `wp_posts` entirely. So this change would also bring the implementation closer to what the issue describes.

properf-wordpress-adapter.php
```diff
@@ -582,6 +636,11 @@ function properf_get_live_data() {
 				'size_bytes'    => 0,
 				'top_size_keys' => array(),
 			),
+			'woo'               => array(
+				'order_items_size_mb'    => 0.0,
+				'order_itemmeta_size_mb' => 0.0,
+				'oldest_order_date'      => null,
+			),
```
pokhiii: Tiny heads-up, not a blocker for this PR — this `'woo' => [...]` default block is the same as the one a few lines down (around line 658), and the `'autoloaded_option'` defaults above are also duplicated in both places.

Issue #11 lists a few more collectors coming after this one (plugin count, hook count, DB sizes), so this pattern will likely keep repeating and the duplication will grow each time.

Might be worth pulling these defaults into a small helper function (e.g. `properf_default_metrics()`) at some point so the next PR only adds one line in one place. Up to you whether to do it in this PR or a later one.

CLAUDE.md
```diff
@@ -0,0 +1,17 @@
+# Claude Code Rules
+
+## Git identity
+- Always use `git config user.name "Ajay-Singh-Adhikari"` and `git config user.email "ajaysinghadhikari697@gmail.com"` before making any commit so both author and committer are set to Ajay's credentials.
```
pokhiii: Curious about this new file — should it be committed to the repo?

A `CLAUDE.md` at the project root becomes a *shared* instruction file. Every developer on this repo who uses Claude Code (a teammate, a future contributor, me reviewing on this checkout) will have these rules applied to them automatically.

This rule in particular caught my eye:

> Always use `git config user.name "Ajay-Singh-Adhikari"` and `git config user.email "ajaysinghadhikari697@gmail.com"` before making any commit so both author and committer are set to Ajay's credentials.

If anyone else runs Claude Code on this checkout, it would try to change their git identity to yours and commit under your name. That would mix up the commit author trail across the repo in a way I don't think you want — commits that should be authored by someone else would end up showing as yours.

For personal Claude Code rules, two common spots keep them personal:

1. `~/.claude/CLAUDE.md` — user-global on your machine, applies to all your projects, doesn't affect anyone else.
2. `CLAUDE.local.md` in the repo root — project-specific, and Claude Code excludes it from git by default.

Could we move this file to one of those two locations instead of committing it here? You can search "Claude Code memory" in the Anthropic docs for the full reference.

(A couple of other rules in this file — about `Co-authored-by:` trailers and what to disclose about AI use in commit messages and PR descriptions — I think are worth a team conversation rather than a personal-config decision. I'll bring those up with you separately on chat.)

Ajay-Singh-Adhikari: You're right, that shouldn't have been committed. Removed CLAUDE.md from the repo and moved the rules to CLAUDE.local.md which is now added to .gitignore so it stays local only.

includes/class-data-collector.php
Ajay-Singh-Adhikari: No, I haven't tested on an HPOS site.
Added the OrderUtil branch to handle both cases — HPOS sites now query wc_orders.date_created_gmt, legacy sites keep the existing wp_posts query. 
I don't currently have access to an HPOS-enabled site to verify the HPOS branch. Happy to test it if you can point me to one.

properf-wordpress-adapter.php
pokhiii: Saw this got addressed in 26b96e2 — `get_default_metrics()` helper is exactly the right shape, and it'll pay off as the next collectors land. Thanks for picking it up. Resolving from my side.

properf-wordpress-adapter.php
pokhiii: Code part is fully in — the archival trigger row and the hardcoded 365 are out, exactly right.

The issue-update ask is still partially open though. I see #12 got a title change and a new bullet about the WP Admin dashboard table, which is good progress. But the body still says "the adapter does not enforce any rule, surfaces raw numbers" and the "What we are missing" table still lists only two signals. The things we agreed in chat — two plugin settings (retention days + table size limit), the AND-condition trigger, the 1-year default, the SankalpTaru reasoning (13 years of orders = 19 MB) — aren't reflected there yet.

Could you fold those into #12's body before the next PR that builds the trigger? And the query-execution-time collector you proposed at the end of the chat needs its own sub-issue under #11 so we don't lose it.

Not blocking this PR's merge — just so the requirements doc catches up with the chat decisions before the next one builds on top.

pokhiii (approved):
LGTM on the code — three raw signals delivered cleanly, the HPOS branch is correct, defaults are deduplicated into a helper, and the CLAUDE.md cleanup is right. Thanks for engaging with the review thoroughly and turning it around quickly.

Approving so this can land. Two follow-ups that don't block this merge but should happen before the next PR builds on top of this:

1. **Update issue #12's body** to reflect the full chat-thread agreement — the two plugin settings (retention days + table size limit), the AND-condition trigger, the 1-year default, and the SankalpTaru reasoning (13 years of orders = 19 MB). Right now the body still describes the old "raw numbers only" scope, which doesn't match where we landed.
2. **File a new sub-issue under #11** for the query-execution-time collector you proposed at the end of our chat, so it's tracked and doesn't get lost.

Nice work overall.

Ajay-Singh-Adhikari:
Both follow-ups done — Issue #12 updated with the agreed direction and #15 for the QET collector was already filed.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #19, opened 2026-05-25
Define plugin release process with a distributable ZIP build


Ajay-Singh-Adhikari:
## Problem

Currently, sharing the plugin means sharing the GitHub repository link. This requires the person deploying to run `composer install` on the server to install dependencies (e.g. Google Cloud BigQuery SDK). If that step is missed, the plugin fails with:

```
Failed: Google SDK not loaded. Check vendor folder.
```

This is not a reliable release process — the person deploying should receive a single, self-contained ZIP that works out of the box.

## Expected Behaviour

A single command should produce a ready-to-install plugin ZIP that includes:
- All plugin PHP files
- `vendor/` (production Composer dependencies)
- No dev-only files (`.git`, `.gitignore`, dev dependencies, etc.)

The person deploying just uploads the ZIP via **WP Admin → Plugins → Add New → Upload Plugin**. No server commands needed.

## Acceptance Criteria

- [ ] A build script exists (e.g. `bin/build.sh`) that generates `build/performance-adapter-wp-{version}.zip`
- [ ] The ZIP includes `vendor/` with production dependencies only (`--no-dev`)
- [ ] The version in the ZIP matches the plugin header version
- [ ] The build output is excluded from git (`.gitignore`)
- [ ] Process is documented so any team member can run a release

PR #20, opened 2026-05-29
feat: add WooCommerce order metrics and QET baseline tracking

### LINKED ISSUE #14: feat: archival signal dashboard — surface order health facts for engineer decision
## Problem

The adapter collects order table sizes and oldest order date (#12). Having the raw numbers is the first step — but the dashboard still can't tell an engineer when archival is actually needed. That decision currently requires manual judgment each time.

## Why an automated trigger doesn't work

Every approach for an automated trigger was explored and dismissed:

- **QET from LIMIT 100 queries** — always fast because MySQL stops at 100 rows. Doesn't change whether the DB is 100MB or 17GB. Not a real signal.
- **Table size threshold** — no universal number works. SankalpTaru has 13 years of orders at 19MB. MFM hit 17.3GB in months. Can't set a fixed number that fits both.
- **Order age only** — permanently ON from day 1 after archival. Archive today, and the very next day the oldest remaining order crosses the retention window again. The condition never turns off.
- **Order age + cooldown** — cooldown is a number picked with no principled basis. A hack on top of a broken approach.
- **Page load derived threshold** — two problems: some platforms can't hit 2s even with a perfectly clean DB, and page load depends on JS, plugins, and external APIs — a slow page doesn't mean the DB is the bottleneck.
- **Per-site relative baseline (30-day)** — if the site is already degraded at install, the baseline is set at the degraded state. The trigger will never fire until QET doubles from an already-bad starting point. Fixing this still requires an engineer to verify the baseline per client, which breaks the tool's core purpose.
- **JOIN query threshold** — the query is a real signal (EXPLAIN confirms it processes rows proportionally as orders accumulate). But setting a threshold against it still fails: the same query on the same data runs at different speeds depending on the server. A site with 100K orders might show 50ms on a managed host and 200ms on shared hosting — not because the DB is more bloated, but because the hardware is weaker.

**Conclusion:** An automated trigger that works universally without human input is unachievable with data available inside WordPress. Every approach either misfires on specific infrastructure or needs context the plugin can't have.

## What we built instead

Surface facts on the dashboard and push them to BigQuery. An engineer looks at these side by side — no guesswork, no missed signals.

| Fact | Status |
|------|--------|
| Oldest order date | ✅ Done |
| Latest order date | ✅ Done |
| Total orders | ✅ Done |
| Count of orders older than N years (configurable retention window) | ✅ Done |
| Last archival date (or never) | ✅ Done |
| Current Query Execution Time | ✅ Done |
| Baseline Query Execution Time | ✅ Done |

**Baseline QET logic:** average of daily QET readings taken after the last archival — locks permanently after 10 post-archival days. If no archival recorded, falls back to average of the 10 lowest readings from the 30-day rolling history. Multiple same-day pushes are averaged into a single daily entry.

## How an engineer uses this

If Current QET is significantly higher than Baseline QET, combined with a high count of orders older than the retention window, the decision is obvious. No automated trigger needed. The plugin removes the need to manually query the DB — that's the real value. The decision stays human.

## Related
- Implemented in: #20
- Raw storage signals: #12
- Query Execution Time collector: #15
- Parent tracking issue: #11

### LINKED ISSUE #15: Collect WooCommerce order query execution time as a performance signal
## Problem

The adapter currently collects order table sizes and oldest order date (#12). These are storage signals — they tell you how much data exists and how old it is. But they don't tell you whether that data is actually slowing down the site right now.

Two sites can have identical table sizes and still have completely different query execution times depending on server specs, indexing, and load. Table size alone is not a reliable performance signal.

## What we need

Measure the actual query execution time (QET) of a representative WooCommerce order query — one that genuinely scales with data volume — and push that number to BigQuery on every scheduled run.

## The correct query to time

For HPOS sites:
```sql
SELECT COUNT(oi.order_item_id)
FROM wp_woocommerce_order_items oi
INNER JOIN wp_wc_orders o ON o.id = oi.order_id
WHERE o.date_created_gmt < DATE_SUB(UTC_TIMESTAMP(), INTERVAL N YEAR)
AND o.status != 'trash'
```

For legacy (non-HPOS) sites:
```sql
SELECT COUNT(oi.order_item_id)
FROM wp_woocommerce_order_items oi
INNER JOIN wp_posts o ON o.ID = oi.order_id
WHERE o.post_date_gmt < DATE_SUB(UTC_TIMESTAMP(), INTERVAL N YEAR)
AND o.post_type = 'shop_order'
AND o.post_status != 'trash'
```

`N` = retention window configured in ProPerf Settings (default: 2 years). The query is filtered by the retention window so it specifically measures the performance cost of the data that would be archived — not the entire dataset. Trashed orders are excluded on both paths so the count reflects active order data only.

**Why this query:** EXPLAIN confirms it processes rows proportionally as orders accumulate — it is a real, environment-specific measure of how the order data volume affects query performance. A LIMIT 100 query was dismissed because MySQL stops at 100 rows regardless of table size and always returns fast.

## What changes

- The adapter runs this query on each scheduled/manual run and records execution time in milliseconds as `woo_query_execution_ms` (INTEGER).
- Multiple same-day pushes are averaged into a single daily history entry to prevent outlier readings from skewing the baseline.
- HPOS and legacy order storage are both handled.
- No threshold logic here — this issue is purely collection. How the signal is used is handled in #14.

## Expected outcome

- ✅ BigQuery receives `woo_query_execution_ms` on every run alongside existing `woo_*` fields.
- ✅ Dashboard shows Current QET and Baseline QET side by side, making degradation visible before it becomes a production issue.

## Related
- Implemented in: #20
- Raw storage signals: #12
- Archival signal dashboard (uses this): #14
- Parent tracking issue: #11

Ajay-Singh-Adhikari:
Closes #14
Closes #15

## What this PR adds

Extends the existing WooCommerce Order Metrics section on the ProPerf dashboard with new data points and QET baseline tracking to support the archival trigger decision.

> **Already in main:** Order Items Table Size, Order Itemmeta Table Size, Oldest Order Date

---

## New metrics added

| Metric | Where it comes from |
|---|---|
| **Latest Order Date** | `MAX()` of order creation date — shows how recently the store was active |
| **Total Orders** | COUNT of all non-trashed orders |
| **Orders Older Than N Years** | COUNT of orders older than the configured retention window. N is now configurable in ProPerf Settings per client (was hardcoded at 2 years) |
| **Last Archival Date** | Set manually in ProPerf Settings after an archival is performed. Resets QET baseline tracking. |
| **Current Query Execution Time (QET)** | Time in ms to execute a JOIN between order items and orders filtered by the retention window — measured on every cron/manual push |
| **Baseline QET** | Daily average QET. Locks after 10 post-archival days. Falls back to avg of 10 lowest readings from 30-day history if no archival recorded. |

---

## How this helps determine if archival is needed

- **Orders Older Than N Years + Total Orders** — shows what percentage of the store is archivable (e.g. 2,420 of 7,688 = 31.5%)
- **QET vs Baseline** — as old order data accumulates, the QET query gets slower. When current QET drifts above the stable baseline, it signals DB performance is degrading and archival should be considered. After archival, the baseline resets from the clean post-archival state.

---

## Other changes

- Retention window wired into all SQL queries (HPOS and legacy both handled)
- "Mark Archival Done" button replaced with **Update Archival Date** link to settings page
- Removed redundant "Oldest Order Age" field
- Same-day QET pushes are averaged into a single daily history entry — prevents one outlier push from skewing the baseline
- Baseline QET label uses plain language: `building new baseline — day 1 of 10` → `stable baseline`
- BigQuery schema fix: `woo_oldest_order_date` marked NULLABLE; `woo_latest_order_date`, `woo_orders_older_than_threshold`, `woo_last_archival_date`, `woo_query_execution_ms`, and `woo_baseline_qet_ms` added as new fields
- PHP compatibility: replaced `str_starts_with()` (PHP 8.0+) with `strpos()` in baseline QET label rendering; replaced `array_key_last()` (PHP 7.3+) with `count($history) - 1` in `record_qet_reading()`
- HPOS QET query: added `AND o.status != 'trash'` filter — was missing, causing inconsistency with the legacy path which already had it
- Timezone fix: `oldest_order_date` and `latest_order_date` now use `strtotime( ... . ' UTC' )` to correctly parse UTC datetime strings from MySQL on non-UTC servers

---

## Pre-merge steps (BigQuery)

For this PR, Ajay is responsible for running the following DDL on the MFM production BigQuery table before merging:

- ADD COLUMN `woo_latest_order_date` DATE
- ADD COLUMN `woo_orders_older_than_threshold` INTEGER
- ADD COLUMN `woo_last_archival_date` DATE
- ADD COLUMN `woo_query_execution_ms` INTEGER
- ADD COLUMN `woo_baseline_qet_ms` INTEGER

After merging, confirm data flows correctly by either triggering a manual push via the ProPerf dashboard or waiting for the next scheduled cron run.


includes/class-data-collector.php
```diff
@@ -102,34 +104,200 @@ public function collect_woo_order_metrics() {
...
+					FROM {$wpdb->prefix}woocommerce_order_items oi
+					INNER JOIN {$wpdb->posts} o ON o.ID = oi.order_id
+					WHERE o.post_date_gmt < DATE_SUB(UTC_TIMESTAMP(), INTERVAL %d YEAR)
+					AND o.post_type = 'shop_order'
+					AND o.post_status != 'trash'",
+					$threshold_years
+				)
 			);
 		}
+		$query_execution_ms = (int) round( ( microtime( true ) - $qet_start ) * 1000 );
```
pokhiii, line 193: Curious about this block when WooCommerce isn't installed on a site.

Today there's no early check for whether WC is active before this method runs. On a site without WC:

- The earlier queries against `wp_posts` (oldest/latest date, total, orders-older-than) all return harmless values — `NULL` or `0` — because no `post_type = 'shop_order'` rows exist. That's defensible.
- But the **QET query here** joins against `{$wpdb->prefix}woocommerce_order_items`. If that table doesn't exist (WC never installed, or just deactivated and tables not yet present), MySQL throws a real error. `$wpdb->get_var()` returns `NULL`, `$wpdb->last_error` gets populated, but **we still record `microtime(true) - $qet_start` as the QET on line 174**. That number is meaningless — it's basically "time MySQL took to refuse the query".

That meaningless reading then enters `properf_qet_history` via `record_qet_reading`, and pollutes the rolling 30-day window and the eventual baseline lock-in.

Two possible fixes:

1. Short-circuit the whole method when WC isn't active — the simplest:
   ```php
   if ( ! function_exists( 'WC' ) ) {
       // return zeroed defaults
   }
   ```
2. Guard just the QET block, since the order-count queries are harmless either way — return `0` for `query_execution_ms` if WC isn't active so the history doesn't get bogus entries.

I'd lean towards (1) — clearer intent — but either works. This was actually a smaller version of the same concern I raised back on PR #13 (`function_exists('WC')` would catch it cheaply); it just matters more now because the consequence is baseline pollution, not just zero values.

includes/class-data-collector.php
```diff
@@ -102,34 +104,200 @@ public function collect_woo_order_metrics() {
...
+		}
+
+		$history = get_option( 'properf_qet_history', array() );
+
+		if ( empty( $history ) ) {
+			return array( 'ms' => null, 'source' => null );
+		}
+
+		if ( $last_archival_date ) {
+			$archival_ts   = strtotime( $last_archival_date );
```
pokhiii: Subtle timezone wart here.

`$last_archival_date` is a `Y-m-d` string (e.g. `'2026-05-15'`), read from the `properf_last_archival_date` option that the engineer set in the WP settings page.

`strtotime('2026-05-15')` parses that as midnight **in the server's local timezone**.

But the history entries' `ts` field is from `time()` — UTC seconds since epoch.

So if the WP server runs in IST (UTC+5:30), `strtotime('2026-05-15')` resolves to `2026-05-14 18:30 UTC`. The `r['ts'] >= $archival_ts` filter on line 260 then ends up including up to ~5.5h of pre-archival readings as post-archival.

In the user's favor (more readings → faster lock-in), so it's not catastrophic — the 10-day lock-in could just trigger slightly earlier than the spec implies, with one almost-pre-archival reading mixed into the baseline.

One-line fix:

```php
$archival_ts = strtotime( $last_archival_date . ' UTC' );
```

Keeps both sides of the comparison in UTC.

properf-wordpress-adapter.php
```diff
@@ -120,15 +122,17 @@ function properf_collect_and_push_metrics() {
...
 
-	$success = $bq_client->push_metrics( $collector->get_data() );
+	$success = $bq_client->push_metrics( $metrics );
 
 	// Persist cron execution info.
 	update_option( 'properf_bq_last_sync', time(), false );
 	update_option( 'properf_bq_last_sync_status', $success ? 'success' : 'error', false );
 
 	if ( $success ) {
+		$collector->record_qet_reading( $metrics['woo']['query_execution_ms'] );
```
pokhiii: Design question — is it intentional that `record_qet_reading` only runs when the BigQuery push succeeds?

Both here (cron) and at line 69 (manual push), the call sits inside the `if ( $success )` branch. So if BigQuery is unhealthy for, say, a week, the local `properf_qet_history` stops growing — and the 10-day post-archival lock-in stalls until BQ comes back.

But the QET reading is a measurement of the **local DB**, completely independent of BigQuery's health. Push is just transmission; the history is local.

Is there a reason to couple the two? Otherwise I'd suggest recording the QET reading unconditionally — call `record_qet_reading` right after `$collector->get_data()`, outside the success branch:

```php
$collector = new ProPerf_Data_Collector();
$metrics   = $collector->get_data();
$collector->record_qet_reading( $metrics['woo']['query_execution_ms'] );

$bq_client = new ProPerf_BigQuery_Client();
$success   = $bq_client->push_metrics( $metrics );
```

That way the baseline keeps building regardless of upload status.

Curious to hear your thinking.

pokhiii (commented):
Big one — closes both #14 (archival signal dashboard) and #15 (QET collector). Read through it twice. The architecture decisions in #14 (no automated trigger, surface facts and let the engineer decide) and the exact JOIN query from #15 are followed faithfully. A few things that stood out as well done:

- HPOS branch handled cleanly across all four query groups (oldest, latest, total, orders-older-than) plus the QET query itself — no mixed schemas.
- Daily-averaging math in `record_qet_reading` is correct: existing avg × count + new reading ÷ (count+1).
- Lock-in logic is right: takes the first 10 post-archival readings (the cleanest ones, just after archival), writes the locked baseline to its own option, and returns it directly on subsequent calls.
- The action hook for the archival date now also `delete_option('properf_baseline_qet_ms')` — that's the right defensive touch, because the previous archival cycle's locked baseline is meaningless after a new archival.
- The dashboard's source-label logic (`stable baseline` / `building new baseline — day N of 10` / `based on N days of data`) reads well — gives the engineer a clear picture of how trustworthy the baseline number is.

Acceptance criteria are met for both issues — every fact in #14's "What we built instead" table is on the dashboard and in BigQuery, and the QET query in #15 matches character-for-character.

Left four inline notes. Three substantive, one nit:

1. QET measurement when WooCommerce isn't installed — pollutes baseline (correctness)
2. Timezone wart in the post-archival filter (minor correctness)
3. QET history only recorded on push success — design question
4. Lost indentation on a dashboard line (cosmetic)

Three things worth surfacing at the PR level too:

### BigQuery schema migration coordination

The deploy note in the description is the right instinct. Worth making more explicit though — the new columns (`woo_orders_older_than_threshold`, `woo_last_archival_date`, `woo_query_execution_ms`, `woo_baseline_qet_ms`) get sent on every push. The load job in `class-bigquery-client.php` uses the schema array in code, but BigQuery itself only auto-adds missing columns when `schema_update_options` contains `ALLOW_FIELD_ADDITION` — which isn't set today. So if the BigQuery table isn't migrated before this PR merges + the cron fires, the next push fails (and every push after until DDL catches up).

Two paths:
- Add `schema_update_options: ['ALLOW_FIELD_ADDITION']` to the load config so BQ auto-adds new columns on the fly. Cheaper but slightly riskier.
- Document a runbook step: run BQ DDL → merge PR → wait for next cron. Manual but explicit.

Which one is the plan?

### PR description mentions a rename that I can't find

> BigQuery schema fixes: `woo_oldest_order_date` marked NULLABLE, `woo_orders_older_than_2y` renamed to `woo_orders_older_than_threshold`

The NULLABLE change I see in the diff. But `woo_orders_older_than_2y` doesn't exist in main's current schema — only the new field name `woo_orders_older_than_threshold` appears (as a fresh addition). Was the rename from a draft branch you didn't push, or am I missing context? Not a blocker — just want to make sure the BigQuery table on the production side doesn't have a stale `_2y` column we need to drop.

### Refactor #23 reminder

The entry file `properf-wordpress-adapter.php` is now **851 lines** after this PR (was 793 before, was 649 before PR #22). This PR is exactly the kind that would have benefited from #23 going first — it touches dashboard render, data collector, BigQuery schema, and the action hook all at once. Each cross-cutting change makes the next one harder to review.

Not blocking this PR. But if #23 lands before the next PR on this area, future iterations get easier. Soft nudge to schedule it.

### Verdict

Holding at `COMMENT`, not approving yet:
- QET-when-WC-absent is a real correctness issue I'd like fixed.
- Want a clear answer on the QET-on-success design coupling and the BigQuery deploy plan.

Nothing else is blocking — mostly polish and one timezone wart. Code-wise this is solid work and the design adherence to #14 / #15 is genuinely good. Looking forward to closing this out.

properf-wordpress-adapter.php
```diff
@@ -222,15 +226,17 @@ function properf_render_dashboard() {
...
-		$oldest_age_days = (int) floor( ( time() - strtotime( $oldest_date ) ) / DAY_IN_SECONDS );
-	}
+	$orders_older_than_threshold = $woo_metrics['orders_older_than_threshold'];
+	$total_orders                = $woo_metrics['total_orders'];
+	$threshold_years             = $woo_metrics['threshold_years'];
+	$last_archival_date   = $woo_metrics['last_archival_date'];
+	$baseline_qet_ms      = $woo_metrics['baseline_qet_ms'];
 
-	$last_sync = get_option( 'properf_bq_last_sync' );
+$last_sync = get_option( 'properf_bq_last_sync' );
```
pokhiii: Tiny nit — looks like the leading tab on this line got lost in a refactor. The line starts at column 0 inside the function body, but everything around it is indented one level.

includes/class-data-collector.php
Ajay-Singh-Adhikari: Good catch. Will add a function_exists('WC') guard at the top of collect_woo_order_metrics() — returning null for all woo metrics when WC isn't active. Will also add a null check in record_qet_reading() so a null QET never enters the history.

includes/class-data-collector.php
Ajay-Singh-Adhikari: Agreed. Will fix with strtotime( $last_archival_date . ' UTC' ) to keep both sides of the comparison in UTC.

properf-wordpress-adapter.php
Ajay-Singh-Adhikari: Makes sense — QET is a local DB measurement and has no dependency on BigQuery's health. Will move `record_qet_reading()` to run right after `get_data()` in both the manual push and cron paths, before the BigQuery client is even initialised.

properf-wordpress-adapter.php
Ajay-Singh-Adhikari: Will fix.

Ajay-Singh-Adhikari:
> Big one — closes both #14 (archival signal dashboard) and #15 (QET collector). Read through it twice. The architecture decisions in #14 (no automated trigger, surface facts and let the engineer decide) and the exact JOIN query from #15 are followed faithfully. A few things that stood out as well done:
> 
> * HPOS branch handled cleanly across all four query groups (oldest, latest, total, orders-older-than) plus the QET query itself — no mixed schemas.
> * Daily-averaging math in `record_qet_reading` is correct: existing avg × count + new reading ÷ (count+1).
> * Lock-in logic is right: takes the first 10 post-archival readings (the cleanest ones, just after archival), writes the locked baseline to its own option, and returns it directly on subsequent calls.
> * The action hook for the archival date now also `delete_option('properf_baseline_qet_ms')` — that's the right defensive touch, because the previous archival cycle's locked baseline is meaningless after a new archival.
> * The dashboard's source-label logic (`stable baseline` / `building new baseline — day N of 10` / `based on N days of data`) reads well — gives the engineer a clear picture of how trustworthy the baseline number is.
> 
> Acceptance criteria are met for both issues — every fact in #14's "What we built instead" table is on the dashboard and in BigQuery, and the QET query in #15 matches character-for-character.
> 
> Left four inline notes. Three substantive, one nit:
> 
> 1. QET measurement when WooCommerce isn't installed — pollutes baseline (correctness)
> 2. Timezone wart in the post-archival filter (minor correctness)
> 3. QET history only recorded on push success — design question
> 4. Lost indentation on a dashboard line (cosmetic)
> 
> Three things worth surfacing at the PR level too:
> 
> ### BigQuery schema migration coordination
> The deploy note in the description is the right instinct. Worth making more explicit though — the new columns (`woo_orders_older_than_threshold`, `woo_last_archival_date`, `woo_query_execution_ms`, `woo_baseline_qet_ms`) get sent on every push. The load job in `class-bigquery-client.php` uses the schema array in code, but BigQuery itself only auto-adds missing columns when `schema_update_options` contains `ALLOW_FIELD_ADDITION` — which isn't set today. So if the BigQuery table isn't migrated before this PR merges + the cron fires, the next push fails (and every push after until DDL catches up).
> 
> Two paths:
> 
> * Add `schema_update_options: ['ALLOW_FIELD_ADDITION']` to the load config so BQ auto-adds new columns on the fly. Cheaper but slightly riskier.
> * Document a runbook step: run BQ DDL → merge PR → wait for next cron. Manual but explicit.
> 
> Which one is the plan?
> 
> ### PR description mentions a rename that I can't find
> > BigQuery schema fixes: `woo_oldest_order_date` marked NULLABLE, `woo_orders_older_than_2y` renamed to `woo_orders_older_than_threshold`
> 
> The NULLABLE change I see in the diff. But `woo_orders_older_than_2y` doesn't exist in main's current schema — only the new field name `woo_orders_older_than_threshold` appears (as a fresh addition). Was the rename from a draft branch you didn't push, or am I missing context? Not a blocker — just want to make sure the BigQuery table on the production side doesn't have a stale `_2y` column we need to drop.
> 
> ### Refactor #23 reminder
> The entry file `properf-wordpress-adapter.php` is now **851 lines** after this PR (was 793 before, was 649 before PR #22). This PR is exactly the kind that would have benefited from #23 going first — it touches dashboard render, data collector, BigQuery schema, and the action hook all at once. Each cross-cutting change makes the next one harder to review.
> 
> Not blocking this PR. But if #23 lands before the next PR on this area, future iterations get easier. Soft nudge to schedule it.
> 
> ### Verdict
> Holding at `COMMENT`, not approving yet:
> 
> * QET-when-WC-absent is a real correctness issue I'd like fixed.
> * Want a clear answer on the QET-on-success design coupling and the BigQuery deploy plan.
> 
> Nothing else is blocking — mostly polish and one timezone wart. Code-wise this is solid work and the design adherence to #14 / #15 is genuinely good. Looking forward to closing this out.

***BigQuery schema migration coordination***

I think we should go with manual DDL — run the column additions on the BigQuery table first, then merge, then let the cron fire. Keeping it manual so test pushes from dev/staging don't auto-pollute the production schema.

***PR description mentions a rename that can't be found***

Confirmed — the production BigQuery table does have woo_orders_older_than_2y as a leftover column. It was added manually during testing. Will drop it via DDL as part of the same migration step before merge.

Ajay-Singh-Adhikari:
***Refactor #23 reminder***

Acknowledged — will schedule it separately after this PR merges. Agree it would have made this PR easier to review if the file were already split. Will prioritise #23 before the next PR that touches the dashboard or data layer.

properf-wordpress-adapter.php
```diff
@@ -333,8 +339,51 @@ function properf_render_dashboard() {
...
+				<tr>
+					<td><strong><?php esc_html_e( 'Last Archival Date', 'properf' ); ?></strong></td>
+					<td>
+						<?php echo $last_archival_date ? esc_html( $last_archival_date ) : esc_html__( 'Never', 'properf' ); ?>
+						<a href="<?php echo esc_url( admin_url( 'admin.php?page=properf-settings' ) ); ?>" class="button button-secondary" style="margin-left: 12px;"><?php esc_html_e( 'Update Archival Date', 'properf' ); ?></a>
+					</td>
+				</tr>
+				<tr>
+					<td><strong><?php esc_html_e( 'Current Query Execution Time', 'properf' ); ?></strong></td>
+					<td><?php echo esc_html( $woo_metrics['query_execution_ms'] . ' ms' ); ?></td>
```
pokhiii, line 363: Regression introduced by the `function_exists('WC')` guard in the collector.

When WooCommerce isn't installed, `collect_woo_order_metrics()` now returns `null` for `query_execution_ms` (collector line 90). This line then evaluates `null . ' ms'`, which PHP coerces to `' ms'` — so the dashboard renders an empty value with just the unit, like this:

> **Current Query Execution Time** &nbsp; &nbsp; ` ms`

The Baseline QET row right below has a proper null-guard (`if ( null !== $baseline_qet_ms )` at line 367) and falls back to *"Not enough data to calculate baseline yet"*. This row needs the same treatment.

Minimal fix:

```php
<td><?php echo null !== $woo_metrics['query_execution_ms']
    ? esc_html( $woo_metrics['query_execution_ms'] . ' ms' )
    : '—'; ?></td>
```

But a few other rows in the WC metrics table have similar half-rendered behavior on a WC-absent site:

- Line 331, 335 (`number_format( null, 2 )` → `0.00 MB`) — misleadingly shows zero size as if the tables existed.
- Line 347, 351 (`number_format( null )` → `0`) — "Total Orders: 0" and "Orders Older Than 2 Years: 0" on a site that has no WooCommerce at all.

These aren't broken the way line 362 is, but they're misleading.

The cleaner approach is probably to **hide the entire "WooCommerce Order Metrics" section when WC isn't installed**, with a single line above the (now-missing) table:

> *WooCommerce is not active on this site.*

That's one place to guard, not seven. And it's truer to the data: when WC is absent, we don't know any of these values, we shouldn't pretend to.

includes/class-data-collector.php
```diff
@@ -102,34 +122,204 @@ public function collect_woo_order_metrics() {
...
+					AND o.post_type = 'shop_order'
+					AND o.post_status != 'trash'",
+					$threshold_years
+				)
+			);
+		}
+		$query_execution_ms = (int) round( ( microtime( true ) - $qet_start ) * 1000 );
+
+		$last_archival_date = get_option( 'properf_last_archival_date', null ) ?: null;
+		$baseline           = $this->get_baseline_qet( $last_archival_date );
```
pokhiii, line 196: Performance concern — not a bug, but worth raising.

`collect_woo_order_metrics()` is called via `get_data()` on every entry point:

- Cron push (once daily) ✓ expected
- Manual "Push to BigQuery" button ✓ expected
- **Every load of the WP Admin ProPerf dashboard page** ← this is the issue

`properf_render_dashboard()` → `properf_get_live_data()` → `$collector->get_data()` → here. So every time an engineer opens the dashboard, the plugin re-runs:

- 6 order-related `COUNT()` / `MIN()` / `MAX()` queries
- The QET JOIN query (designed to scale with order volume — that's the whole point of #15)
- Plus 2 reads of non-autoloaded options inside `get_baseline_qet()`

On a large store the QET JOIN alone could take several seconds. So opening the dashboard makes the dashboard slow — which is awkward for a plugin meant to diagnose slow sites.

One approach: cache the most recent metrics snapshot.

```php
$cached = get_transient( 'properf_woo_metrics_snapshot' );
if ( false !== $cached ) {
    return $cached;
}
// ... compute ...
set_transient( 'properf_woo_metrics_snapshot', $result, HOUR_IN_SECONDS );
```

The cron and the manual "Push to BigQuery" button can force a fresh measurement (delete the transient before computing) so QET history isn't built from cache. The dashboard view reads the cache and shows the most recent values.

Alternative shape: have the dashboard read the *last pushed* snapshot from a stored option instead of recomputing. Either works.

Not blocking this PR, but worth a follow-up issue at minimum so the cost doesn't compound silently.

pokhiii (commented):
Final-state review (not commit-by-commit).

## Process

All four of my earlier inline threads were resolved by you, not by me. In a code review the **reviewer** resolves threads once they have verified the fix — the developer's role is to push the fix and reply on the thread. Self-resolving removes my ability to use the open-threads list as a verification checklist, and on this PR it hid a regression. Going forward as a working agreement: you reply, I resolve. I've re-opened the four threads, replied on each with the per-thread verdict, and resolved the three that are actually done.

## Inline notes (where the per-line concerns live)

- `properf-wordpress-adapter.php:362` — render regression introduced by the WC-absent guard fix.
- `includes/class-data-collector.php:168-195` — performance concern: the heavy QET JOIN runs on every dashboard page load, not just on cron / manual push.

## Two items still unanswered from my first review

**a. BigQuery schema migration coordination.** The load job in `class-bigquery-client.php:148` doesn't set `schemaUpdateOptions: ['ALLOW_FIELD_ADDITION']`. If the BigQuery table isn't DDL-migrated before merge + next cron, every push will fail until the table catches up. Either set that option so BQ auto-adds the new columns, or document a runbook step (`BQ DDL → merge → wait for next cron`). The deploy note in the PR description names the columns but doesn't say who runs the DDL or when.

**b. PR description claims `woo_orders_older_than_2y` was renamed.** That field doesn't exist in `main`'s schema — `woo_orders_older_than_threshold` is a fresh addition. Is there a stale column on the production BigQuery table that needs a manual drop? If yes, that's a runbook step; if no, the PR description needs correcting.

## WordPress best-practices audit

Clean:

- SQL injection — all dynamic values go through `$wpdb->prepare()`, verified line by line.
- Output escaping — all new dashboard output uses `esc_html` / `esc_url` / `esc_attr`.
- Capability check (`current_user_can('manage_options')`) + nonce (`check_admin_referer`) on the push.
- Translation strings + text domain `properf` consistent.
- `properf_qet_history` and `properf_baseline_qet_ms` set with `autoload=false` — no autoload-cache bloat.
- HPOS detection — `class_exists` + capability check, graceful fallback.

Worth raising:

- **Race condition in `record_qet_reading()`** — two concurrent pushes (cron mid-run + manual click) could both read `properf_qet_history`, both append, and one overwrites. WP options don't have atomic update primitives. Low probability today (cron is daily, manual pushes rare) — worth a one-line awareness note in the docblock so a future maintainer doesn't get bitten if cron frequency is raised.
- **External dependency footprint** — plugin loads the full Google Cloud BigQuery SDK from `vendor/`. Pre-existing. Relevant when #19 (distributable ZIP) lands — could revisit by switching to a thin HTTP client against the BigQuery REST API. Not urgent.
- **`error_log()` usage** in the cron handler is unconditional. WP convention is to gate behind `WP_DEBUG_LOG`. Pre-existing.

## Acceptance criteria

- #14 — every fact in the "What we built instead" table is on the dashboard and in BigQuery ✅
- #15 — the JOIN query (HPOS and legacy variants) matches the spec character-for-character; recorded as `woo_query_execution_ms INTEGER` ✅

includes/class-data-collector.php
pokhiii: Holding this one open.

The collector-side fix is correct — `function_exists('WC')` guard at collector line 79 cleanly returns null for all woo metrics, and `record_qet_reading()` skips on `null === $qet_ms`. Good defence in depth.

But the change introduced a render regression on the **dashboard side** that the self-resolve missed: when WC is absent, `query_execution_ms` is `null`, and the dashboard at `properf-wordpress-adapter.php:362` concatenates that into `null . ' ms'` — which PHP coerces to `' ms'`. So the Current QET row renders as a blank value with just the unit. I left a fresh inline note on line 362 with two fix options (null-guard that one row, or — cleaner — hide the entire WooCommerce Order Metrics section when WC isn't installed, since several other rows in it also render misleadingly with zeros).

Resolving from my side once the consumer side is fixed too.

includes/class-data-collector.php
pokhiii: Verified at collector line 277 — `strtotime( $last_archival_date . ' UTC' )` keeps both sides of the post-archival filter in UTC. Exactly the fix. Resolving from my side.

properf-wordpress-adapter.php
pokhiii: Verified — `record_qet_reading()` is now called right after `$collector->get_data()` in both the manual push (adapter line 60) and the cron path (line 126), before the BigQuery push. Local QET history grows regardless of upload state. Resolving from my side.

properf-wordpress-adapter.php
pokhiii: Verified — `$last_sync = get_option( 'properf_bq_last_sync' );` at line 239 now has the correct leading tab. Resolving from my side.

properf-wordpress-adapter.php
Ajay-Singh-Adhikari: Went with the cleaner approach — When WC is absent, the whole table is hidden and replaced with "WooCommerce is not active on this site."

includes/class-data-collector.php
Ajay-Singh-Adhikari: Valid concern. Will open a follow-up issue to track the transient caching approach — out of scope for this PR but shouldn't get lost.

[Created a new Issue](https://github.com/ColoredCow/performance-adapter-wp/issues/24)

includes/class-data-collector.php
Ajay-Singh-Adhikari: Dashboard side is fixed too — the entire WooCommerce Order Metrics section is now hidden when WC isn't active (properf-wordpress-adapter.php:321). Both sides are guarded.

Ajay-Singh-Adhikari:
> Final-state review (not commit-by-commit).
> 
> ## Process
> All four of my earlier inline threads were resolved by you, not by me. In a code review the **reviewer** resolves threads once they have verified the fix — the developer's role is to push the fix and reply on the thread. Self-resolving removes my ability to use the open-threads list as a verification checklist, and on this PR it hid a regression. Going forward as a working agreement: you reply, I resolve. I've re-opened the four threads, replied on each with the per-thread verdict, and resolved the three that are actually done.
> 
> ## Inline notes (where the per-line concerns live)
> * `properf-wordpress-adapter.php:362` — render regression introduced by the WC-absent guard fix.
> * `includes/class-data-collector.php:168-195` — performance concern: the heavy QET JOIN runs on every dashboard page load, not just on cron / manual push.
> 
> ## Two items still unanswered from my first review
> **a. BigQuery schema migration coordination.** The load job in `class-bigquery-client.php:148` doesn't set `schemaUpdateOptions: ['ALLOW_FIELD_ADDITION']`. If the BigQuery table isn't DDL-migrated before merge + next cron, every push will fail until the table catches up. Either set that option so BQ auto-adds the new columns, or document a runbook step (`BQ DDL → merge → wait for next cron`). The deploy note in the PR description names the columns but doesn't say who runs the DDL or when.
> 
> **b. PR description claims `woo_orders_older_than_2y` was renamed.** That field doesn't exist in `main`'s schema — `woo_orders_older_than_threshold` is a fresh addition. Is there a stale column on the production BigQuery table that needs a manual drop? If yes, that's a runbook step; if no, the PR description needs correcting.
> 
> ## WordPress best-practices audit
> Clean:
> 
> * SQL injection — all dynamic values go through `$wpdb->prepare()`, verified line by line.
> * Output escaping — all new dashboard output uses `esc_html` / `esc_url` / `esc_attr`.
> * Capability check (`current_user_can('manage_options')`) + nonce (`check_admin_referer`) on the push.
> * Translation strings + text domain `properf` consistent.
> * `properf_qet_history` and `properf_baseline_qet_ms` set with `autoload=false` — no autoload-cache bloat.
> * HPOS detection — `class_exists` + capability check, graceful fallback.
> 
> Worth raising:
> 
> * **Race condition in `record_qet_reading()`** — two concurrent pushes (cron mid-run + manual click) could both read `properf_qet_history`, both append, and one overwrites. WP options don't have atomic update primitives. Low probability today (cron is daily, manual pushes rare) — worth a one-line awareness note in the docblock so a future maintainer doesn't get bitten if cron frequency is raised.
> * **External dependency footprint** — plugin loads the full Google Cloud BigQuery SDK from `vendor/`. Pre-existing. Relevant when #19 (distributable ZIP) lands — could revisit by switching to a thin HTTP client against the BigQuery REST API. Not urgent.
> * **`error_log()` usage** in the cron handler is unconditional. WP convention is to gate behind `WP_DEBUG_LOG`. Pre-existing.
> 
> ## Acceptance criteria
> * #14 — every fact in the "What we built instead" table is on the dashboard and in BigQuery ✅
> * #15 — the JOIN query (HPOS and legacy variants) matches the spec character-for-character; recorded as `woo_query_execution_ms INTEGER` ✅

**a. BigQuery schema migration coordination**

Proceeding with manual DDL — intentional to prevent schema changes from dev/staging from auto-propagating to the production table. PR description updated to document the required DDL steps and who runs them before merge.

**b. `woo_orders_older_than_2y` rename claim**

The rename note was based on a field that existed only in the test BigQuery table during development — not present on any client's Bigquery Table. PR description corrected: rename reference removed, required schema changes listed explicitly.

Ajay-Singh-Adhikari:
Also pushed three additional correctness fixes found during a self-review:

- PHP 7.x compatibility: replaced `str_starts_with()` with `strpos()` in baseline QET label rendering — would have been a fatal on PHP < 8.0
- HPOS QET query: added `AND o.status != 'trash'` — was missing, inconsistent with the legacy path which already had it
- Timezone fix: `oldest_order_date` and `latest_order_date` now use `strtotime(... . ' UTC')` — same pattern as the `last_archival_date` fix from the previous round
- Also replaced `array_key_last()` with `count($history) - 1` — `array_key_last()` requires **PHP 7.3+**.

includes/class-data-collector.php
pokhiii: Verified — `properf-wordpress-adapter.php:321` now wraps the entire WooCommerce Order Metrics table in `if ( function_exists( 'WC' ) )` with an else branch showing *"WooCommerce is not active on this site."* (line 392). Both sides are guarded now: collector returns null, renderer doesn't reach the broken rows. Resolving from my side.

properf-wordpress-adapter.php
pokhiii: Verified — the `function_exists('WC')` guard at adapter line 321 hides the whole table when WC is absent, so the `' ms'` render path doesn't get reached. Cleaner than per-row null-guarding, exactly the approach I was hoping for. Resolving from my side.

includes/class-data-collector.php
pokhiii: Tracked in #24. Resolving from my side.

pokhiii (approved):
All open asks from my prior pass are closed:

- The `' ms'` render regression: fixed by hiding the entire WooCommerce Order Metrics section when WC isn't active. Cleaner UX than per-row null-guards.
- Performance concern: tracked in #24 as a follow-up.
- Race-condition docblock on `record_qet_reading()`: added with the exact framing from the audit.
- Two unanswered items from the first review are now answered in the PR description — the new "Pre-merge steps (BigQuery)" section names the DDL columns and assigns ownership, and the `woo_orders_older_than_2y` mention is gone (corrected to "added as new fields").

A few additional self-review catches in this round worth calling out — independent improvements that strengthen the PR:

- HPOS QET query was missing `AND o.status != 'trash'`. Now matches the legacy path.
- `woo_latest_order_date` was being collected but missing from the BigQuery schema array. Caught and added.
- PHP 7.x compatibility: `str_starts_with()` → `strpos()`, `array_key_last()` → `count() - 1`. Both would have been fatal on PHP < 8.0.
- Timezone fix extended to `oldest_order_date` and `latest_order_date` formatting (`strtotime( ... . ' UTC' )`).

Good self-review discipline on those.

Approving.

PR #22, opened 2026-05-29
feat: add configurable retention window and last archival date to plugin settings

### LINKED ISSUE #21: Add configurable retention window and optional last archival date to plugin settings
## Problem

Different clients have different archival policies. One client may want to archive orders older than 1 year, another after 3 years. Right now the plugin uses a fixed value of 2 years everywhere — in the SQL queries, in the dashboard labels, in the metrics pushed to BigQuery. There is no way to change this without modifying the plugin code directly.

Relying on a hardcoded number means the tool is not client-specific. The data it shows and the queries it runs may not match the client's actual archival policy at all.

### 1. Retention window setting (required)

A new field in the plugin settings page where the engineer sets the retention window for that client — for example, 2 years or 3 years.

Once set, the plugin uses this value everywhere:
- The SQL query that counts old orders uses the configured retention window instead of a fixed number
- The QET query also uses it
- The dashboard label updates to reflect it — e.g. "Orders Older Than 3 Years" instead of always saying 2
- The metric pushed to BigQuery reflects the correct window

Default value: **2 years** (so existing behaviour is preserved if no value is set).

### 2. Last archival date field (optional)

An optional date field in the plugin settings where the engineer can record when the last archival was performed for that client. This is purely informational — it gives a clearer picture on the dashboard when reviewing order health alongside QET and order count.

If the engineer does not fill it in, the field stays blank and nothing breaks.

**QET history reset on date change:** When the archival date is updated, the stored QET history (`properf_qet_history`) should be reset so the baseline recomputes from the new post-archival state. This reset should be triggered via the `update_option_properf_last_archival_date` action hook — not inside a sanitize_callback — so it fires regardless of how the option is updated (settings form, CLI, programmatic call).

> Note: `properf_qet_history` is created by the QET collector tracked in #15. This feature has a dependency on #15.

## What changes after this

- The tool becomes client-specific. Each client's WordPress admin reflects their own retention policy.
- The dashboard and BigQuery data are accurate to what that client actually cares about.
- Engineers no longer need to touch code to adjust the retention window for a new client.

## Acceptance Criteria

- [ ] Retention window field exists on the plugin settings page, default value 2, max value 20
- [ ] Invalid input preserves the previously saved value rather than resetting to default
- [ ] SQL queries use the configured retention window value
- [ ] Dashboard label reflects the configured value
- [ ] Last archival date is an optional date field on the settings page
- [ ] Changing the archival date resets `properf_qet_history` via `update_option_properf_last_archival_date` hook
- [ ] Both values are saved and used by the plugin correctly

## Related
- #14 — archival signal dashboard
- #15 — QET collector (dependency for QET history reset)

Ajay-Singh-Adhikari:
Closes #21

## What changes

Adds a new **WooCommerce Settings** section to the plugin settings page with two fields:

### Retention Window (Years)
A number input that sets how many years back the plugin should look when counting archival candidates. Defaults to 2, max 20. This PR adds the settings UI only — SQL queries, dashboard labels, and BigQuery wiring will follow in a separate PR.

### Last Archival Date (Optional)
A date input where the engineer can record when orders were last archived for this client. Supports backdating for clients where archival happened before the plugin was installed. When this date is changed, the QET history resets so the baseline recomputes from the new post-archival state.

### DIFF
--- a/properf-wordpress-adapter.php
+++ b/properf-wordpress-adapter.php
@@ -568,12 +568,14 @@ function properf_bigquery_private_key_field() {
  */
 function properf_sanitize_threshold_years( $value ) {
 	$value = intval( $value );
-	return $value >= 1 ? $value : 2;
+	if ( $value < 1 || $value > 20 ) {
+		return intval( get_option( 'properf_archival_threshold_years', 2 ) );
+	}
+	return $value;
 }
 
 /**
  * Sanitize archival date — must be a valid Y-m-d date or empty.
- * Resets QET history when the date changes so baseline recomputes from the new post-archival state.
  *
  * @param string $value Submitted value.
  * @return string Sanitized date or empty string.
@@ -587,11 +589,25 @@ function properf_sanitize_archival_date( $value ) {
 	if ( ! $date || $date->format( 'Y-m-d' ) !== $value ) {
 		return get_option( 'properf_last_archival_date', '' );
 	}
-	if ( $value !== get_option( 'properf_last_archival_date', '' ) ) {
-		update_option( 'properf_qet_history', array(), false );
-	}
 	return $value;
 }
+
+/**
+ * Reset QET history when archival date changes so baseline recomputes from the new post-archival state.
+ *
+ * @param string $old Previous value.
+ * @param string $new New value.
+ */
+add_action(
+	'update_option_properf_last_archival_date',
+	function ( $old, $new ) {
+		if ( $old !== $new ) {
+			update_option( 'properf_qet_history', array(), false );
+		}
+	},
+	10,
+	2
+);
 
 /**
  * WooCommerce section callback.
@@ -606,7 +622,7 @@ function properf_woo_section_callback() {
 function properf_archival_threshold_years_field() {
 	$value = intval( get_option( 'properf_archival_threshold_years', 2 ) );
 	?>
-	<input type="number" name="properf_archival_threshold_years" value="<?php echo esc_attr( $value ); ?>" min="1" step="1" class="small-text" />
+	<input type="number" name="properf_archival_threshold_years" value="<?php echo esc_attr( $value ); ?>" min="1" max="20" step="1" class="small-text" />
 	<p class="description"><?php esc_html_e( 'Orders older than this many years will be counted as archival candidates. Default is 2.', 'properf' ); ?></p>
 	<?php
 }
--- a/properf-wordpress-adapter.php
+++ b/properf-wordpress-adapter.php
@@ -442,6 +442,49 @@ function properf_register_settings() {
 		'properf-settings',
 		'properf_bigquery_section'
 	);
+
+	register_setting(
+		'properf_bigquery_settings',
+		'properf_archival_threshold_years',
+		array(
+			'type'              => 'integer',
+			'sanitize_callback' => 'properf_sanitize_threshold_years',
+			'default'           => 2,
+		)
+	);
+
+	register_setting(
+		'properf_bigquery_settings',
+		'properf_last_archival_date',
+		array(
+			'type'              => 'string',
+			'sanitize_callback' => 'properf_sanitize_archival_date',
+			'default'           => '',
+		)
+	);
+
+	add_settings_section(
+		'properf_woo_section',
+		__( 'WooCommerce Settings', 'properf' ),
+		'properf_woo_section_callback',
+		'properf-settings'
+	);
+
+	add_settings_field(
+		'properf_archival_threshold_years',
+		__( 'Retention Window (Years)', 'properf' ),
+		'properf_archival_threshold_years_field',
+		'properf-settings',
+		'properf_woo_section'
+	);
+
+	add_settings_field(
+		'properf_last_archival_date',
+		__( 'Last Archival Date', 'properf' ),
+		'properf_last_archival_date_field',
+		'properf-settings',
+		'properf_woo_section'
+	);
 }
 
 /**
@@ -518,6 +561,68 @@ function properf_bigquery_private_key_field() {
 }
 
 /**
+ * Sanitize retention window — must be a positive integer, min 1.
+ *
+ * @param mixed $value Submitted value.
+ * @return int Sanitized value.
+ */
+function properf_sanitize_threshold_years( $value ) {
+	$value = intval( $value );
+	return $value >= 1 ? $value : 2;
+}
+
+/**
+ * Sanitize archival date — must be a valid Y-m-d date or empty.
+ * Resets QET history when the date changes so baseline recomputes from the new post-archival state.
+ *
+ * @param string $value Submitted value.
+ * @return string Sanitized date or empty string.
+ */
+function properf_sanitize_archival_date( $value ) {
+	$value = sanitize_text_field( $value );
+	if ( '' === $value ) {
+		return '';
+	}
+	$date = \DateTime::createFromFormat( 'Y-m-d', $value );
+	if ( ! $date || $date->format( 'Y-m-d' ) !== $value ) {
+		return get_option( 'properf_last_archival_date', '' );
+	}
+	if ( $value !== get_option( 'properf_last_archival_date', '' ) ) {
+		update_option( 'properf_qet_history', array(), false );
+	}
+	return $value;
+}
+
+/**
+ * WooCommerce section callback.
+ */
+function properf_woo_section_callback() {
+	echo '<p>' . esc_html__( 'Configure WooCommerce-specific settings for this client.', 'properf' ) . '</p>';
+}
+
+/**
+ * Retention window field.
+ */
+function properf_archival_threshold_years_field() {
+	$value = intval( get_option( 'properf_archival_threshold_years', 2 ) );
+	?>
+	<input type="number" name="properf_archival_threshold_years" value="<?php echo esc_attr( $value ); ?>" min="1" step="1" class="small-text" />
+	<p class="description"><?php esc_html_e( 'Orders older than this many years will be counted as archival candidates. Default is 2.', 'properf' ); ?></p>
+	<?php
+}
+
+/**
+ * Last archival date field.
+ */
+function properf_last_archival_date_field() {
+	$value = get_option( 'properf_last_archival_date', '' );
+	?>
+	<input type="date" name="properf_last_archival_date" value="<?php echo esc_attr( $value ); ?>" class="regular-text" />
+	<p class="description"><?php esc_html_e( 'Optional. Enter the date when orders were last archived for this client. Used as a reference point on the dashboard.', 'properf' ); ?></p>
+	<?php
+}
+
+/**
  * Render Settings page.
  */
 function properf_render_settings() {
@@ -579,6 +684,29 @@ function properf_render_settings() {
 						</th>
 						<td>
 							<?php properf_bigquery_private_key_field(); ?>
+						</td>
+					</tr>
+				</tbody>
+			</table>
+
+			<h2><?php esc_html_e( 'WooCommerce Settings', 'properf' ); ?></h2>
+			<?php properf_woo_section_callback(); ?>
+			<table class="form-table" role="presentation">
+				<tbody>
+					<tr>
+						<th scope="row">
+							<label for="properf_archival_threshold_years"><?php esc_html_e( 'Retention Window (Years)', 'properf' ); ?></label>
+						</th>
+						<td>
+							<?php properf_archival_threshold_years_field(); ?>
+						</td>
+					</tr>
+					<tr>
+						<th scope="row">
+							<label for="properf_last_archival_date"><?php esc_html_e( 'Last Archival Date', 'properf' ); ?></label>
+						</th>
+						<td>
+							<?php properf_last_archival_date_field(); ?>
 						</td>
 					</tr>
 				</tbody>


properf-wordpress-adapter.php
```diff
@@ -517,6 +560,68 @@ function properf_bigquery_private_key_field() {
...
+	if ( '' === $value ) {
+		return '';
+	}
+	$date = \DateTime::createFromFormat( 'Y-m-d', $value );
+	if ( ! $date || $date->format( 'Y-m-d' ) !== $value ) {
+		return get_option( 'properf_last_archival_date', '' );
+	}
+	if ( $value !== get_option( 'properf_last_archival_date', '' ) ) {
+		update_option( 'properf_qet_history', array(), false );
+	}
```
pokhiii: Two curious questions about this block.

**1. Where did the QET-history reset behavior come from?**

Issue #21 describes the Last Archival Date as *"purely informational — it gives a clearer picture on the dashboard when reviewing order health"* and *"if the engineer does not fill it in, the field stays blank and nothing breaks"*. The QET-history reset is new behavior not in the spec. Same pattern we hit on PR #13 — chat-thread agreement that didn't make it into the issue. If the reset is the right behavior, could we add it to #21 (or to #15, since QET history lives there) so the spec catches up with the code?

Also — `properf_qet_history` is the option created by the QET collector tracked in #15, which is still open. So we're resetting an option that doesn't exist yet. Calling `update_option` on a missing key is harmless in WP (it just creates it), but it suggests this PR has a dependency on #15 that isn't acknowledged.

**2. Even if the reset is the right behavior, is `sanitize_callback` the right place for it?**

A sanitize_callback is meant to be pure — input in, sanitized value out. Doing an `update_option` side effect inside it has two subtle risks:

- WordPress can invoke `sanitize_callback` more than once per save (validation passes). The reset would fire each time. Today it's idempotent (clearing twice = clearing once), but that's brittle.
- The reset only fires when the date changes **via the Settings form**. If anything else updates the option — a future CLI command, a migration, a programmatic `update_option('properf_last_archival_date', ...)` call — `sanitize_callback` is bypassed, so the QET history would silently get stale and the baseline would be wrong.

A cleaner spot is the `update_option_properf_last_archival_date` action hook — fires whenever the option actually changes, no matter the code path:

```php
add_action( 'update_option_properf_last_archival_date', function ( $old, $new ) {
    if ( $old !== $new ) {
        update_option( 'properf_qet_history', array(), false );
    }
}, 10, 2 );
```

Reading: https://developer.wordpress.org/reference/hooks/update_option_option/

Curious to hear your thinking on both of these.

properf-wordpress-adapter.php
```diff
@@ -517,6 +560,68 @@ function properf_bigquery_private_key_field() {
...
 
+/**
+ * Sanitize retention window — must be a positive integer, min 1.
+ *
+ * @param mixed $value Submitted value.
+ * @return int Sanitized value.
+ */
+function properf_sanitize_threshold_years( $value ) {
+	$value = intval( $value );
+	return $value >= 1 ? $value : 2;
```
pokhiii: Two small thoughts on this sanitizer:

**1. Loses the user's existing value on invalid input.**

The date sanitizer right below (line 588) preserves the previous value when input is invalid: `return get_option( 'properf_last_archival_date', '' );`. This one resets to `2` instead. So if a user has `5` saved and accidentally submits `0` or empty, they silently lose their `5` and get reset to the default.

Could mirror the date pattern for consistency:

```php
return $value >= 1 ? $value : intval( get_option( 'properf_archival_threshold_years', 2 ) );
```

That way the fallback is "keep what you had", not "reset to default".

**2. No upper bound.**

A user could submit `9999` and it would save fine. Not a security concern, just a sanity check for typos. Two ways to add a ceiling: `max="20"` on the `<input>` (browser-side, can be bypassed) plus a clamp in the sanitizer like:

```php
$value = min( $value, 20 );
```

Neither of these is blocking — small UX polish.

pokhiii (commented):
Thanks for getting the settings UI up — the WP Settings API is wired correctly (`register_setting` + `sanitize_callback` + defaults), and the field rendering follows the existing BigQuery section pattern, which is nice and consistent.

One main thing I want to talk through though.

The PR description says:

> Previously this was hardcoded to 2 years everywhere — in SQL queries, dashboard labels, and BigQuery metrics. Now it reads from this setting, making the plugin client-specific.

But when I look at the diff and search the codebase, I don't see any SQL query, dashboard label, or BigQuery code that reads from these new settings. They get registered, sanitized, saved to options, and shown on the page — but nothing else in the plugin uses them yet. I also couldn't find any existing "2 years" hardcoded value anywhere to be replaced (the current collector just returns `MIN(date_created)`, no year filter).

Mapping that to issue #21's acceptance criteria:

- [x] Retention window field exists on the settings page, default 2
- [ ] **SQL queries use the configured retention window value** — I don't see this happening
- [ ] **Dashboard label reflects the configured value** — I don't see this happening
- [x] Last archival date is an optional date field on the settings page
- [ ] Both values are saved and **used by the plugin correctly** — saved yes, used not yet

Two ways this could go:

1. **Settings UI was meant to be a separate, smaller PR**, and the SQL / dashboard / BigQuery wiring is coming in a follow-up. Totally fine — settings UI alone is a clean unit of work. In that case, could you update the PR description to say "settings UI only, wiring follows in a separate PR"? And it might be worth splitting issue #21 into two issues (one for the UI, one for the wiring) so the acceptance criteria match what's actually shipping in each PR.

2. **The wiring is supposed to be in this PR**, and it just got missed. In that case it needs to be added before merge.

Which is it?

One other small thing worth surfacing: issue #12's body (the "Agreed direction" section) still says *"Order retention window (days) — default: 365"*, while issue #21 and this PR use **years, default 2**. The two issues disagree on the unit. Worth picking one and updating the other so future readers don't get confused.

Left a couple of inline notes on smaller things below.

properf-wordpress-adapter.php
Ajay-Singh-Adhikari: Good catch. The update_option call inside the sanitize callback was wrong — sanitize callbacks can run during register_setting as well, not just on form submit, so side effects there are unreliable. Moved the QET history reset to an update_option_properf_last_archival_date action hook instead. That fires only when the value actually changes through a real update_option call.

properf-wordpress-adapter.php
Ajay-Singh-Adhikari: Fair point. Updated the sanitizer to fall back to the currently saved value instead of hardcoding 2, so an existing setting isn't silently overwritten on invalid input. Also added max="20" on the input and a matching clamp in the sanitizer.

properf-wordpress-adapter.php
pokhiii: Nice. The action hook approach is exactly right, and the `$old !== $new` guard inside it keeps it idempotent. Also a good observation in your reply that sanitize_callback can fire during `register_setting` registration too — that's a subtle WP behavior worth knowing.

Resolved from my side.

properf-wordpress-adapter.php
pokhiii: Looks good. `[1, 20]` clamp + fall back to the saved value matches the date sanitizer pattern cleanly, and the `max="20"` on the input gives the browser-level guard too. Thanks for picking both up.

Resolved from my side.

pokhiii (commented):
Re-reviewed at `a23f433`. Inline fixes are clean — both threads addressed with the right patterns:

- Date sanitizer is now pure; QET reset moved to the `update_option_properf_last_archival_date` action hook with an `$old !== $new` guard. Exactly the shape I had in mind. Bonus that the stale "Resets QET history" line in the docblock was cleaned up too.
- Years sanitizer falls back to the saved value, clamps to `[1, 20]`, and the `<input>` got `max="20"`. Consistent with how the date sanitizer handles invalid input.
- Issue #21 was updated with the new ACs (max 20, preserve-on-invalid, QET reset via the hook) and a note about the #15 dependency. Good — the spec now matches the code, which was the main process ask.

Two things from my first review are still open though, that I don't see addressed yet:

### 1. The wiring question

The PR description still says:

> Previously this was hardcoded to 2 years everywhere — in SQL queries, dashboard labels, and BigQuery metrics. Now it reads from this setting, making the plugin client-specific.

But I still don't see any SQL query, dashboard label, or BigQuery code that reads from `properf_archival_threshold_years`. Searched the whole codebase — the only callers are the field renderer (which just displays the value back in the form) and the sanitizer's fallback. Same for `properf_last_archival_date`.

Mapping that to issue #21's current AC list:

- [x] Retention window field exists, default 2, max 20
- [x] Invalid input preserves the previously saved value
- [ ] **SQL queries use the configured retention window value**
- [ ] **Dashboard label reflects the configured value**
- [x] Last archival date is an optional date field
- [x] Changing the archival date resets `properf_qet_history` via `update_option_*` hook
- [ ] **Both values are saved and used by the plugin correctly** — saved yes, used not yet

So 5/7 met, 2/7 not (effectively 3 with the catch-all). The central question I asked in my first summary is still unanswered: is this PR intentionally settings-UI-only, with the SQL / dashboard / BigQuery wiring in a follow-up PR? Or did the wiring get missed?

I'd be totally happy with "settings UI only" as the answer — it's a clean unit. But in that case, two small things to close the loop:

1. Update this PR's description so it doesn't overpromise — something like *"Adds the settings UI for retention window and last archival date. SQL / dashboard / BigQuery wiring to follow in a separate PR."*
2. Either split #21 into a UI issue (closed by this PR) and a wiring issue (open as follow-up), or leave #21 open after this merges with the wiring ACs still unchecked. Whichever feels right.

### 2. Issue #12 unit drift

Issue #12's body (the "Agreed direction for the archival trigger" section) still says:

> **Order retention window (days)** — default: 365.

While issue #21 and this PR use **years, default 2**. The two issues disagree on the unit, which will confuse future readers. Could you update #12 to match (years/2), or pick one unit and update the other? Same kind of doc-vs-code hygiene as last time.

---

Not calling this APPROVE yet just because of the wiring question — once you confirm the intent (UI-only follow-up or wiring-in-this-PR), and the description + #21 / #12 are aligned with whichever it is, this is good to merge. Code-wise nothing else is blocking from my side.

Ajay-Singh-Adhikari:
> Re-reviewed at `a23f433`. Inline fixes are clean — both threads addressed with the right patterns:
> 
> * Date sanitizer is now pure; QET reset moved to the `update_option_properf_last_archival_date` action hook with an `$old !== $new` guard. Exactly the shape I had in mind. Bonus that the stale "Resets QET history" line in the docblock was cleaned up too.
> * Years sanitizer falls back to the saved value, clamps to `[1, 20]`, and the `<input>` got `max="20"`. Consistent with how the date sanitizer handles invalid input.
> * Issue #21 was updated with the new ACs (max 20, preserve-on-invalid, QET reset via the hook) and a note about the #15 dependency. Good — the spec now matches the code, which was the main process ask.
> 
> Two things from my first review are still open though, that I don't see addressed yet:
> 
> ### 1. The wiring question
> The PR description still says:
> 
> > Previously this was hardcoded to 2 years everywhere — in SQL queries, dashboard labels, and BigQuery metrics. Now it reads from this setting, making the plugin client-specific.
> 
> But I still don't see any SQL query, dashboard label, or BigQuery code that reads from `properf_archival_threshold_years`. Searched the whole codebase — the only callers are the field renderer (which just displays the value back in the form) and the sanitizer's fallback. Same for `properf_last_archival_date`.
> 
> Mapping that to issue #21's current AC list:
> 
> * [x]  Retention window field exists, default 2, max 20
> * [x]  Invalid input preserves the previously saved value
> * [ ]  **SQL queries use the configured retention window value**
> * [ ]  **Dashboard label reflects the configured value**
> * [x]  Last archival date is an optional date field
> * [x]  Changing the archival date resets `properf_qet_history` via `update_option_*` hook
> * [ ]  **Both values are saved and used by the plugin correctly** — saved yes, used not yet
> 
> So 5/7 met, 2/7 not (effectively 3 with the catch-all). The central question I asked in my first summary is still unanswered: is this PR intentionally settings-UI-only, with the SQL / dashboard / BigQuery wiring in a follow-up PR? Or did the wiring get missed?
> 
> I'd be totally happy with "settings UI only" as the answer — it's a clean unit. But in that case, two small things to close the loop:
> 
> 1. Update this PR's description so it doesn't overpromise — something like _"Adds the settings UI for retention window and last archival date. SQL / dashboard / BigQuery wiring to follow in a separate PR."_
> 2. Either split #21 into a UI issue (closed by this PR) and a wiring issue (open as follow-up), or leave #21 open after this merges with the wiring ACs still unchecked. Whichever feels right.
> 
> ### 2. Issue #12 unit drift
> Issue #12's body (the "Agreed direction for the archival trigger" section) still says:
> 
> > **Order retention window (days)** — default: 365.
> 
> While issue #21 and this PR use **years, default 2**. The two issues disagree on the unit, which will confuse future readers. Could you update #12 to match (years/2), or pick one unit and update the other? Same kind of doc-vs-code hygiene as last time.
> 
> Not calling this APPROVE yet just because of the wiring question — once you confirm the intent (UI-only follow-up or wiring-in-this-PR), and the description + #21 / #12 are aligned with whichever it is, this is good to merge. Code-wise nothing else is blocking from my side.

## 1. Wiring question

This PR is **limited to the settings UI only.**

While working on [Issue #14](https://github.com/coloredcow/performance-adapter-wp/issues/14), I noticed the dashboard needs to surface the number of orders older than the retention window. That made it clear we first need a way for clients to define the retention window through plugin settings — before wiring it into the SQL queries, dashboard labels, and BigQuery payload.

The SQL query wiring (using `properf_archival_threshold_years` in the `INTERVAL` clause) and the dashboard label update are part of `feat/woo-qet-collector`, not this PR.

The PR description was inaccurate — it implied the wiring was already done, which it isn't. I'll update it to:

> Adds the settings UI for retention window and last archival date. SQL, dashboard, and BigQuery wiring will follow in a separate PR.

For [Issue #21](https://github.com/coloredcow/performance-adapter-wp/issues/21), I'll keep it open after this merges and leave the two wiring-related ACs unchecked so the follow-up work remains clearly tracked.

---

## 2. Issue #12 – Unit drift

The unit changed from days to years during the design discussion — years are more intuitive for archival and retention cycles.

I'll update [Issue #12](https://github.com/coloredcow/performance-adapter-wp/issues/12) to reflect years as the unit with a default of 2.

pokhiii (approved):
Thanks for the clear scope clarification and the quick doc updates. Confirming what's now in place:

- ✅ PR description is corrected — accurately describes this PR as settings-UI-only with the wiring deferred.
- ✅ Issue #12 body updated — retention window unit is now `years / default 2`, harmonized with #21 and this PR. No more unit drift between the two issues.
- ✅ Plan for #21 confirmed in your comment: keep open after merge with the three wiring ACs (SQL queries use the value, dashboard label reflects it, "values used by the plugin correctly") unchecked, so the follow-up work stays clearly tracked.
- ✅ Reasoning for the split out of #14 makes sense — building the settings dependency before the consumer that needs it is the right call. The earlier PR description just hadn't caught up with the scope.

Code-wise (at `a23f433`) everything is clean:
- Sanitizers are pure; QET reset lives on the `update_option_*` action hook with an `$old !== $new` guard.
- Years sanitizer clamps `[1, 20]` and falls back to saved value on invalid input.
- `max="20"` on the input element.
- Issue #21 spec matches the code.

Approving. Nice work on this PR — both rounds of feedback were addressed thoroughly and the engagement on the threads was good.

---

One orthogonal thing worth flagging, **not blocking this merge**: I filed #23 yesterday about the size of `properf-wordpress-adapter.php` — it's now 793 lines after this PR and holds ~67% of all plugin PHP. It's pre-existing tech debt, not something this PR introduced, but the file will grow further with the wiring PR, #14, and #15 all touching the same admin/settings code. If you have appetite, doing #23 before the wiring follow-up lands would mean those next PRs go into a cleaner structure. No expectation either way — just on your radar.

PR #26, opened 2026-06-09
Refactor: split properf-wordpress-adapter.php into focused class files

### LINKED ISSUE #23: Refactor: split properf-wordpress-adapter.php into focused modules under includes/
## Problem

The main plugin file `properf-wordpress-adapter.php` is now **793 lines** and growing fast — up from **649** before PR #22 (+144 lines, ~22% growth in a single PR). It currently holds ~**67%** of all PHP code in the plugin (793 of ~1170 lines across the codebase).

It mixes at least 8 distinct concerns:

1. Plugin bootstrap and constants
2. WP Admin dashboard render (`properf_render_dashboard`)
3. Settings registration (`register_setting`, `add_settings_section`, `add_settings_field` calls)
4. Field renderer functions — currently 7, one per setting
5. Sanitize callbacks — one per setting that needs validation
6. Settings page render (`properf_render_settings`)
7. Action hooks (`update_option_properf_last_archival_date` listener)
8. Live-data fallback (`properf_get_live_data`)

## Why it matters

**Each new setting currently requires 5 separate additions in this one file:**

- `register_setting()` call
- `add_settings_field()` call
- A sanitize function
- A field renderer function
- A hand-coded `<tr>` row inside `properf_render_settings()`

That last one is because the settings page hand-codes its HTML instead of calling `do_settings_sections('properf-settings')`. As a side effect, the `add_settings_section()` / `add_settings_field()` registrations in `properf_register_settings()` are effectively decorative — they are registered but never rendered through the WP Settings API. Every new setting carries 5 separate touches in one file when it should carry 3.

With more settings already planned (table size limit per #14, retention-window wiring still pending under #21, QET-related settings under #15), this pattern compounds with every PR.

**`includes/` already exists** for `class-bigquery-client.php` and `class-data-collector.php`. The convention for separating concerns is in place — admin / settings / dashboard concerns just never got pulled out.

## Proposed structure

Candidate split, all under `includes/`:

- **`class-admin-settings.php`** — owns all settings registration, field renderers, sanitize callbacks, and the settings page render. Should switch to `do_settings_sections()` so each new setting only needs `register_setting()` + `add_settings_field()` + sanitize fn + renderer, with no per-setting `<tr>` boilerplate.
- **`class-admin-dashboard.php`** — owns the WP Admin dashboard render: the Autoloaded Options table, the WooCommerce Order Metrics table, the "Push to BigQuery" action.
- **`class-live-data.php`** — owns `properf_get_live_data()` and the default-metrics helper.

The entry file `properf-wordpress-adapter.php` shrinks back to plugin bootstrap, autoloader wiring, and the activation / deactivation / cron hooks. Target: ~150 lines.

## Acceptance criteria

- [ ] Each concern listed under "Problem" lives in its own class file under `includes/`
- [ ] `properf-wordpress-adapter.php` is under 200 lines and contains only plugin bootstrap and lifecycle hooks
- [ ] Settings page uses `do_settings_sections()` — no hand-coded `<tr>` per setting; adding a new setting requires only register + add_settings_field + sanitize fn + renderer (no extra render-function touch)
- [ ] All existing behavior preserved — settings save and load correctly, dashboard renders identically, BigQuery push works, the `update_option_properf_last_archival_date` hook still fires
- [ ] Existing feature work in flight (#14, #15, the wiring PR for #21) can land cleanly on top of the new structure

## Why now

Two recent PRs (#13: +120 lines; #22: +144 lines) added admin/settings code to this file. Three more PRs touching the same area are in flight or imminent (#14, #15, and the wiring follow-up for #21). If we don't restructure before those land, the file will likely cross 1200 lines and the refactor cost will be 3× higher than today.

This is a **refactor-only** issue — no behavior changes, no new features.

## Related

- #11 — tracking issue for server-side metrics (each new metric tends to add settings + dashboard code)
- #14, #15, #21 — open feature work that will touch the same file
- PRs #13 and #22 — where the growth happened

Ajay-Singh-Adhikari:
Closes #23

## What changed

`properf-wordpress-adapter.php` was 855 lines mixing 8 concerns. This PR splits it into focused class files under `includes/` and reduces the main file to 113 lines. Several correctness and standards issues in the original code are fixed as part of the split.

### New files

| File | Responsibility |
|---|---|
| `includes/class-admin-dashboard.php` | Admin menu registration, dashboard page render, BigQuery push handler, settings-saved notice handler |
| `includes/class-admin-settings.php` | Settings registration (options, sections, fields), field renderers, sanitize callbacks, settings page render, QET history reset on archival date change |
| `includes/class-live-data.php` | Live data retrieval from collector, safe fallback on error |
| `assets/css/admin.css` | Admin styles extracted from inline `style=""` attributes |

### `class-data-collector.php` — `collect_and_push()` added

The manual push handler and the daily cron were duplicating the same pipeline: collect → record QET → push → persist sync status. `collect_and_push()` centralises the shared work and returns `{ success, error }` so each caller keeps only its own notification (transient vs `error_log`).

### `properf-wordpress-adapter.php` after refactor

Contains only:
- Plugin header, constants, `require_once` wiring
- Activation / deactivation hooks
- Cron scheduler + cron push handler (delegates to `collect_and_push()`)
- `properf_get_next_midnight()` utility

---

### Security

- **Capability check added to `handle_bigquery_push()`** — the original checked the nonce but not `current_user_can('manage_options')`. Nonce verifies request origin; capability verifies permission. Both are required independently.
- **`sanitize_key()` on superglobal reads** — `$_GET['page']` and `$_POST['option_page']` now go through `sanitize_key()` before comparison.

### Hook registration

- **`update_option_properf_last_archival_date` hook moved outside `is_admin()` guard** — the original anonymous closure registered this hook only during admin requests. WP-CLI and cron calls to `update_option()` would silently skip QET history reset. Now registered via `register_persistent_hooks()` which is called unconditionally from `properf_init()`.
- **`wp_redirect` filter registered admin-only** — previously registered globally on every request including frontend checkout flows. Now only registers when `is_admin()` is true.
- **`handle_settings_notices()` scoped to ProPerf pages** — transient is only read on `properf` or `properf-settings` pages, not on every admin page load.
- **Anonymous hook replaced with named static method** — `reset_qet_on_archival_change()` makes the hook removable.

### Post/Redirect/Get for BigQuery push

Previously `properf_handle_bigquery_push()` called `add_settings_error()` inline in the POST handler and re-rendered the dashboard in the same request. Refreshing after a push resubmitted the form.

`handle_bigquery_push()` now stores the result in a per-user transient (60s TTL) via `push_notice_key()`, then calls `wp_safe_redirect()` + `exit`. On the following GET, `handle_settings_notices()` reads the transient and passes it to `add_settings_error()`.

### Type contract

The original `count` field in `autoloaded_option` held either an `int` or an error string depending on the code path, protected only by a silent `is_numeric()` guard in the template. Errors now live in a dedicated `error` key (`null` in the happy path, string on error); `count` is always `int`. The `error` key is present in every code path — both `collect_autoloaded_options()` and the fallback in `ProPerf_Live_Data` — so callers can rely on a consistent shape.

### Exception handling

`collect_and_push()` wraps `$this->get_data()` in a try-catch. A DB exception during data collection previously propagated uncaught, silently killing the cron job. Now it returns `['success' => false, 'error' => $message]` so the cron handler logs the failure and the manual push shows the error notice.

### Structure

- **`register_settings()` split into three private helpers** — `register_options()`, `register_sections()`, `register_fields()`. Adding a new setting now has a clear home for each concern.
- **`format_baseline_label()` extracted** — the 20-line `if/elseif/else` that parsed `baseline_qet_source` strings and built `$source_label` was inline in the render method. Extracted to a private static helper.
- **`push_notice_key()` extracted** — the transient key literal `'properf_push_notice_' . get_current_user_id()` appeared 4 times across two methods. Centralised in a private helper.
- **Private helpers placed after public methods** — in both `ProPerf_Admin_Dashboard` and `ProPerf_Admin_Settings`, private static helpers appear at the bottom of the class so the public interface is visible first when reading the file.

### Settings page

`properf_render_settings()` previously hand-coded two `<table>` blocks with a `<tr>` per field. The `add_settings_section()` / `add_settings_field()` registrations existed but were rendered manually rather than through WP.

`ProPerf_Admin_Settings::render_settings()` now uses `do_settings_sections('properf-settings')`. Adding a new setting requires only `register_setting()` + `add_settings_field()` + sanitize callback + field renderer — no per-setting `<tr>` in the render function.

### Output escaping and i18n

- `wp_die()` uses `esc_html__()` in both dashboard and settings classes
- `printf()` replaced with `echo esc_html( sprintf() )` in all KB/MB size outputs
- `add_settings_error()` message arguments use `esc_html__()`
- All user-facing strings translated: `'Unauthorized'`, `'Never'`, push notices, button labels, error fallback strings in `ProPerf_Live_Data`
- `id` attributes added to all settings form fields (required for `<label>` associations)

### Bug fixes

- `(float)` cast added to `$gmt_offset` in `timezone_name_from_abbr()` — prevents integer math errors on half-hour timezone offsets (e.g. IST +5:30)
- Dead code removed from `properf_get_next_midnight()`: `$tz = $tz ? $tz : 'UTC'` made the `gmt_offset` fallback block unreachable
- `reset_qet_on_archival_change()` guards with `! empty( $old )` so it only fires on actual date change, not on first-ever save

---

## Acceptance criteria

- [x] Each concern lives in its own class file under `includes/`
- [x] `properf-wordpress-adapter.php` is under 200 lines (113 lines)
- [x] Settings page uses `do_settings_sections()` — no hand-coded `<tr>` per setting
- [x] BigQuery push uses Post/Redirect/Get — no form resubmission on refresh
- [x] `update_option_properf_last_archival_date` hook fires in all contexts (admin, CLI, cron)
- [x] Admin hooks only register in admin context (`is_admin()` guard on `properf_init()`)
- [x] Nonce and capability check both present on `handle_bigquery_push()`
- [x] Shared push pipeline in `collect_and_push()` — single source of truth for cron and manual push
- [x] DB exception in `collect_and_push()` returns structured error instead of uncaught propagation
- [x] `autoloaded_option.count` is always `int`; `error` key present in all code paths (`null` when none)
- [x] Private helpers follow public methods in each class file
- [x] All existing behaviour preserved — settings save/load, dashboard renders identically, BigQuery push works


includes/class-admin-settings.php
```diff
@@ -0,0 +1,373 @@
...
+ */
+class ProPerf_Admin_Settings {
+
+	/**
+	 * Register hooks.
+	 */
+	public static function init() {
+		add_action( 'admin_init', array( __CLASS__, 'register_settings' ), 20 );
+		add_filter( 'wp_redirect', array( __CLASS__, 'settings_redirect' ) );
+		add_action( 'update_option_properf_last_archival_date', array( __CLASS__, 'reset_qet_on_archival_change' ), 10, 2 );
```
pokhiii: **Real regression from the `is_admin()` refactor — worth fixing before merge.**

Before this PR, the `update_option_properf_last_archival_date` action hook was registered at file-load time in the main plugin file, so it fired in **every** context — admin form save, WP-CLI, REST, cron, programmatic option update.

After this PR, the hook is registered inside `ProPerf_Admin_Settings::init()`, which is only invoked when `is_admin()` is true (main file line 49). So the QET-reset behavior now fires only on the admin settings form path.

The specific case this breaks: an engineer runs `wp option update properf_last_archival_date 2026-01-15` from WP-CLI after a manual archival. `is_admin()` is false in CLI context, so the hook isn't registered, so `properf_qet_history` and `properf_baseline_qet_ms` don't get reset. The locked baseline stays bound to the previous archival cycle until something else triggers a reset — but nothing else does. Silent staleness, hard to debug.

The fix splits the class init into two: persistent hooks always register; admin-only hooks gate on `is_admin()`. Suggested shape:

```php
/**
 * Register hooks that must fire in any context (CLI, REST, cron, frontend).
 */
public static function register_persistent_hooks() {
    add_action(
        'update_option_properf_last_archival_date',
        array( __CLASS__, 'reset_qet_on_archival_change' ),
        10,
        2
    );
}

/**
 * Register admin-only hooks.
 */
public static function init() {
    add_action( 'admin_init', array( __CLASS__, 'register_settings' ), 20 );
    add_filter( 'wp_redirect', array( __CLASS__, 'settings_redirect' ) );
}
```

And in `properf-wordpress-adapter.php:properf_init()`:

```php
function properf_init() {
    ProPerf_Admin_Settings::register_persistent_hooks();
    if ( is_admin() ) {
        ProPerf_Admin_Dashboard::init();
        ProPerf_Admin_Settings::init();
    }
}
```

The other admin hooks (`register_settings`, `settings_redirect`) are correctly admin-only — only the `update_option_*` listener needs to escape the gate.

includes/class-admin-dashboard.php
```diff
@@ -0,0 +1,332 @@
...
+			'manage_options',
+			'properf-settings',
+			array( 'ProPerf_Admin_Settings', 'render_settings' )
+		);
+	}
+
+	/**
+	 * Handle manual BigQuery push form submission.
+	 */
+	public static function handle_bigquery_push() {
```
pokhiii, line 75: Defense-in-depth nit.

`handle_bigquery_push()` verifies the nonce via `check_admin_referer()`, which confirms (a) the nonce matches and (b) the user is logged in. But it doesn't verify the user actually has `manage_options` — only that they had a valid nonce at some point. In practice the form is only rendered on the dashboard (which IS gated by `current_user_can('manage_options')` at `render_dashboard:153`), so realistically only admins ever obtain a usable nonce. So the risk is low.

But the pattern in WP core / VIP guidelines is to capability-check before any state-changing action, even when a nonce is present — they're orthogonal protections (nonce = CSRF, capability = authorisation). A subscriber whose nonce was somehow exposed (e.g., via a logged debug dump) could currently trigger a push.

One-line addition right after the nonce check:

```php
check_admin_referer( 'properf_push_action', 'properf_push_nonce' );
if ( ! current_user_can( 'manage_options' ) ) {
    wp_die( esc_html__( 'Unauthorized', 'properf' ) );
}
```

Not blocking by itself, but cheap and matches the discipline already shown on the render functions.

pokhiii (commented):
Final-state review of #23's refactor.

Genuinely strong refactor. The structural split is clean, behavior is mostly preserved, and a lot of bonus polish landed alongside the move. Quick reality-check on the issue #23 ACs:

- ✅ Each concern lives in its own class file under `includes/` — dashboard, settings, live-data
- ✅ Main file under 200 lines (123 lines)
- ✅ Settings page uses `do_settings_sections()` — verified line 353; new settings need no `<tr>` boilerplate in render
- ✅ All five existing settings + the two from PR #22 routed through the WP Settings API properly
- ✅ Existing in-flight feature work (#14, #15, #21 wiring) has a clear home

Bonus improvements worth calling out, since they go beyond #23's scope:

- **Post / Redirect / Get on the BigQuery push** — refreshing after a push no longer re-submits the form. Transient-based notice carries through.
- **`is_admin()` guard on `properf_init`** — admin classes don't load on frontend requests anymore. Real perf win.
- **Anonymous closure for QET reset replaced with named static method** — hook is now removable via `remove_action`. Good thinking.
- **`! empty( $old )` guard on `reset_qet_on_archival_change`** — first-ever save no longer triggers a meaningless reset. Subtle but correct.
- **Output escaping cleanup** — `esc_html( sprintf(...) )` replaces unescaped `printf` everywhere; `esc_textarea` for the private key field; all user-facing strings now in `__()` / `esc_html__()`.
- **CSS extracted** to `assets/css/admin.css`, enqueued only on ProPerf admin pages via the hook check on line 33.
- **Sanitizer errors surface to the user** — `add_settings_error` from inside `sanitize_threshold_years` and `sanitize_archival_date` so the engineer sees "previous value restored" instead of silent fallback. Nice UX detail.

## What needs fixing

Two inline notes below. One is a real regression from the `is_admin()` move; one is defense-in-depth.

## WP best-practices audit (final-state)

Clean:
- **Capability checks** on both render functions (`render_dashboard:153`, `render_settings:343`). ✅
- **Nonces** on push form + verified before processing (`check_admin_referer` at dashboard line 79). ✅
- **Output escaping**: every dynamic value goes through an `esc_*` helper. ✅
- **i18n / text domain `properf`** consistent across all new files, including error fallbacks in `ProPerf_Live_Data`. ✅
- **Hook removability** — every callback uses named methods, no anonymous closures. ✅
- **Settings API** used properly now — `register_setting`, `add_settings_section`, `add_settings_field`, `do_settings_sections`. ✅
- **Static method dispatch** (`array( __CLASS__, 'method' )`) — keeps hook registrations introspectable. ✅
- **`wp_safe_redirect`** instead of raw `wp_redirect` — same-host enforced. ✅
- **Asset enqueueing scoped to ProPerf pages** via hook check at line 33 — no global stylesheet load. ✅
- **No new SQL** in this PR (pure refactor), existing `$wpdb->prepare()` calls intact. ✅

Minor nits (low priority, not blocking):
- `$_GET['page']` (settings:110) and `$_POST['option_page']` (settings:368) are read without `sanitize_text_field()` / `sanitize_key()`. Both end up in strict equality checks against literal values so the practical risk is zero, but WPCS will flag these.
- `error_log()` calls in the cron handler (main file lines 86, 92) remain unconditional — pre-existing, not introduced here.

## No regressions found in (verified line by line)

- Cron registration, deactivation cleanup ✅
- BigQuery push success/failure paths and error message storage ✅
- Settings save → redirect → notice display (PRG roundtrip) ✅
- Dashboard render of all metric rows (WC-active and WC-absent branches) ✅
- The `update_option_properf_last_archival_date` action fires when invoked through the settings form ✅ — *but does not fire from non-admin contexts; see inline note*

This is a textbook refactor PR — clear scope, behavior-preserving with one fixable exception, and material code-quality wins as side effects. Once the two inline items are addressed, this is approve-ready.

includes/class-admin-settings.php
Ajay-Singh-Adhikari: Good catch — I missed that the update_option_* hook was only landing in admin context after the refactor. The WP-CLI case makes it obvious: running wp option update properf_last_archival_date after a manual archival is exactly the kind of workflow we'd use, and silently leaving the locked baseline stale would be confusing to debug later.

Went with the register_persistent_hooks() split you suggested — persistent hooks called unconditionally in properf_init(), admin-only hooks inside is_admin(). Clean separation and makes the intent explicit in the code.

includes/class-admin-dashboard.php
Ajay-Singh-Adhikari: Makes sense — I'd been thinking of the nonce as covering both CSRF and access control, but you're right that they're orthogonal. The nonce proves the request is legitimate; the capability check proves the user is allowed to act on it. A subscriber who somehow ends up with a valid nonce (debug log, shared session) shouldn't be able to trigger a push.

Added current_user_can('manage_options') right after the nonce check. One line, cheap, consistent with how render_dashboard() already gates access.

includes/class-admin-settings.php
pokhiii: Verified at `class-admin-settings.php:20-22` and `properf-wordpress-adapter.php:49`. `register_persistent_hooks()` is now called unconditionally from `properf_init()` — the `update_option_*` listener fires regardless of context. `init()` keeps the admin-only hooks (`register_settings`, `settings_redirect`) inside the `is_admin()` gate. Exactly the split. Resolving from my side.

includes/class-admin-dashboard.php
pokhiii: Verified at `class-admin-dashboard.php:80-82`. `current_user_can( 'manage_options' )` check sits right after `check_admin_referer()`, with `wp_die( esc_html__( 'Unauthorized', 'properf' ) )` on failure. Nonce + capability are now orthogonal protections. Resolving from my side.

pokhiii (approved):
Both inline items addressed and verified — see per-thread replies.

Four above-and-beyond improvements landed alongside, worth calling out:

- **`$_GET['page']` sanitized** with `sanitize_key()` at `handle_settings_notices:100` — addresses the WPCS nit I'd noted in passing.
- **Shared push pipeline** — `ProPerf_Data_Collector::collect_and_push()` extracts the duplicated cron + manual-push logic into one method with consistent exception handling and a `{success, error}` result shape. Cron handler in the main file shrinks to 7 lines.
- **Count type contract fix** — `'count'` in the autoload metrics is now always numeric, and error messages live in a separate `'error'` field. Dashboard renders the error if present, else `number_format(count)`. Cleaner contract than overloading the same field as int-or-string.
- **`register_settings()` split** into `register_options()` / `register_sections()` / `register_fields()`, plus `push_notice_key()` and `format_baseline_label()` private helpers. Each method now does one thing.

All concerns from the first pass closed; no regressions found in the new code paths.

Approving.

PR #29, opened 2026-06-11
Cache WooCommerce metrics snapshot on dashboard load

### LINKED ISSUE #24: Cache WooCommerce metrics snapshot on dashboard load
## Overview

Every load of the ProPerf dashboard re-runs all WooCommerce order queries, including the QET JOIN query that is intentionally designed to scale with order volume. On a large store this makes the dashboard itself slow — which is counterproductive for a plugin meant to diagnose slow sites.

The fix is to serve the dashboard from a cached snapshot and only run live queries during cron and manual push.

**In scope:** Cache the metrics snapshot for dashboard reads.
**Out of scope:** Changes to how QET history or baseline are calculated.

## Requirements

1. **Dashboard reads from cache**
   - Opening the ProPerf dashboard does not re-run WC order queries if a cached snapshot exists.
   - Cache TTL: 1 hour (`HOUR_IN_SECONDS`).

2. **Cron and manual push always compute fresh**
   - Both paths delete the transient before calling `get_data()` so QET history is always built from a live measurement, never from cache.

3. **No cache bleed into QET history**
   - `record_qet_reading()` is only called from the cron and manual push paths, which already bypass the cache.

## Acceptance Criteria

- [ ] Opening the dashboard a second time within an hour does not re-run the WC order queries (verify via query log or `SAVEQUERIES`).
- [ ] Clicking "Push to BigQuery" forces a fresh computation and updates the cache.
- [ ] Cron push forces a fresh computation and updates the cache.
- [ ] QET history entries are only recorded from live measurements.
- [ ] Dashboard display is unchanged — same metrics, same format.
- [ ] Existing non-WC metrics on the dashboard are unaffected (no regression).

## Peer Dev QA Checklist

- [ ] Acceptance Criteria of the User Story/Bug
- [ ] Admin and non-Admin flows
- [ ] Multi-tenant / multi-organization users
- [ ] Text wrapping in cards/containers
- [ ] For new UI changes, please test at the following resolutions:
  - [ ] 1920x1080
  - [ ] 1366x768
- [ ] Test features sensitive to larger volumes of data

## Dev Notes

The bottleneck is `collect_woo_order_metrics()` in `includes/class-data-collector.php:78`, called on every dashboard render via:

`properf_render_dashboard()` → `properf_get_live_data()` (`properf-wordpress-adapter.php:839`) → `$collector->get_data()` (line 848)

Proposed approach — add to `collect_woo_order_metrics()`:

```php
$cached = get_transient( 'properf_woo_metrics_snapshot' );
if ( false !== $cached ) {
    return $cached;
}
// ... existing compute logic ...
set_transient( 'properf_woo_metrics_snapshot', $result, HOUR_IN_SECONDS );
```

In both the manual push path (`properf-wordpress-adapter.php:59`) and the cron path (line 125), delete the transient before calling `get_data()`:

```php
delete_transient( 'properf_woo_metrics_snapshot' );
$metrics = $collector->get_data();
```

Alternative: have the dashboard read the last pushed snapshot from a stored WP option instead of a transient. Either works.

### From dev — please validate before picking up

- Call chain confirmed in code: `properf_get_live_data()` at line 839 calls `$collector->get_data()` at line 848; dashboard calls this at line 222.
- Manual push calls `get_data()` at line 59; cron at line 125.
- Verify transient key `properf_woo_metrics_snapshot` doesn't conflict with any existing transient in the codebase before using it.

### Prerequisites

1. WordPress install with WooCommerce active and at least some order data.
2. Enable `SAVEQUERIES` or use Query Monitor plugin to observe DB queries.

### Dashboard caching

1. Open the ProPerf dashboard — note queries run.
2. Open the ProPerf dashboard again within 1 hour.
3. Confirm: the WC order queries do not re-run on the second load.

### Fresh measurement on push

1. Open dashboard (warms cache).
2. Click "Push to BigQuery."
3. Confirm: `collect_woo_order_metrics()` ran a live query (check QET history updated with a new entry).
4. Reload dashboard — confirm cached values reflect the fresh push.

### Regression check

1. Confirm all dashboard metrics still display correctly after the change.
2. Confirm baseline QET calculation is unaffected.

## Related

- Raised in: #20 (PR review)
- Relates to: #15

Ajay-Singh-Adhikari:
Closes #24

## Problem

Every ProPerf dashboard load re-ran the full set of WooCommerce order queries — table size lookups, date range scans, COUNT queries, and the QET JOIN — even when the data hadn't changed. On large databases this adds unnecessary DB load on every page view.

## Solution

Cache the output of `collect_woo_order_metrics()` as a 1-hour WordPress transient. The dashboard hits the cache on repeated loads; only pushes (manual or cron) run live queries.

## Changes

**`includes/class-data-collector.php`**
- Add `WOO_METRICS_CACHE_KEY` constant as the single source of truth for the transient key.
- `collect_woo_order_metrics()`: return cached transient on hit; compute and cache on miss (1-hour TTL via `HOUR_IN_SECONDS`).
- `collect_and_push()`: delete transient before collecting so pushes always use fresh data; after `record_qet_reading()`, call `refresh_cached_baseline()` and update `$metrics` before the BigQuery push to keep the BigQuery row and dashboard baseline consistent.
- Add `refresh_cached_baseline()` (private): after a push records a new QET reading, rewrites only the baseline fields in the cached snapshot so the dashboard immediately reflects the post-push baseline without expiring the rest of the snapshot.
- Add `bust_metrics_cache()` (public static): owns the `delete_transient` call so settings hooks delegate to the class that owns the cache.

**`includes/class-admin-settings.php`**
- `register_persistent_hooks()`: add `update_option_` and `add_option_` hooks for both `properf_last_archival_date` and `properf_archival_threshold_years` to call `ProPerf_Data_Collector::bust_metrics_cache()`. Both hooks are needed because `update_option_` does not fire on the first-ever save of an option.

### DIFF
--- a/includes/class-data-collector.php
+++ b/includes/class-data-collector.php
@@ -270,16 +270,20 @@ class ProPerf_Data_Collector {
 	/**
 	 * Refresh the baseline fields in the cached snapshot after a new QET reading is recorded.
 	 * Keeps the rest of the snapshot intact so the dashboard reflects the post-push baseline immediately.
+	 * Returns the updated baseline array, or null if no snapshot was cached.
+	 *
+	 * @return array|null {ms: int|null, source: string|null}, or null if no cached snapshot exists.
 	 */
 	private function refresh_cached_baseline() {
 		$cached = get_transient( 'properf_woo_metrics_snapshot' );
 		if ( false === $cached ) {
-			return;
+			return null;
 		}
-		$baseline                          = $this->get_baseline_qet( $cached['woo']['last_archival_date'] );
+		$baseline = $this->get_baseline_qet( $cached['woo']['last_archival_date'] );
 		$cached['woo']['baseline_qet_ms']     = $baseline['ms'];
 		$cached['woo']['baseline_qet_source'] = $baseline['source'];
 		set_transient( 'properf_woo_metrics_snapshot', $cached, HOUR_IN_SECONDS );
+		return $baseline;
 	}
 
 	/**
@@ -301,7 +305,11 @@ class ProPerf_Data_Collector {
 		}
 
 		$this->record_qet_reading( $metrics['woo']['query_execution_ms'] );
-		$this->refresh_cached_baseline();
+		$baseline = $this->refresh_cached_baseline();
+		if ( null !== $baseline ) {
+			$metrics['woo']['baseline_qet_ms']     = $baseline['ms'];
+			$metrics['woo']['baseline_qet_source'] = $baseline['source'];
+		}
 		$bq_client = new ProPerf_BigQuery_Client();
 		$success   = $bq_client->push_metrics( $metrics );
 
--- a/includes/class-data-collector.php
+++ b/includes/class-data-collector.php
@@ -95,7 +95,7 @@ class ProPerf_Data_Collector {
 			);
 		}
 
-		// Cache is invalidated by collect_and_push() and by settings changes via ProPerf_Admin_Settings::bust_woo_metrics_cache().
+		// Cached for 1 hour; invalidated on push and on relevant settings changes.
 		$cached = get_transient( 'properf_woo_metrics_snapshot' );
 		if ( false !== $cached ) {
 			return $cached;
@@ -277,7 +277,7 @@ class ProPerf_Data_Collector {
 			return;
 		}
 		$baseline                          = $this->get_baseline_qet( $cached['woo']['last_archival_date'] );
-		$cached['woo']['baseline_qet_ms']  = $baseline['ms'];
+		$cached['woo']['baseline_qet_ms']     = $baseline['ms'];
 		$cached['woo']['baseline_qet_source'] = $baseline['source'];
 		set_transient( 'properf_woo_metrics_snapshot', $cached, HOUR_IN_SECONDS );
 	}
--- a/includes/class-data-collector.php
+++ b/includes/class-data-collector.php
@@ -95,6 +95,11 @@ class ProPerf_Data_Collector {
 			);
 		}
 
+		$cached = get_transient( 'properf_woo_metrics_snapshot' );
+		if ( false !== $cached ) {
+			return $cached;
+		}
+
 		global $wpdb;
 
 		$db_name  = DB_NAME;
@@ -196,7 +201,7 @@ class ProPerf_Data_Collector {
 		$last_archival_date = get_option( 'properf_last_archival_date', null ) ?: null;
 		$baseline           = $this->get_baseline_qet( $last_archival_date );
 
-		return array(
+		$result = array(
 			'woo' => array(
 				'order_items_size_mb'    => $items_size ? round( floatval( $items_size ), 4 ) : 0.0,
 				'order_itemmeta_size_mb' => $itemmeta_size ? round( floatval( $itemmeta_size ), 4 ) : 0.0,
@@ -211,6 +216,10 @@ class ProPerf_Data_Collector {
 				'baseline_qet_source'    => $baseline['source'],
 			),
 		);
+
+		set_transient( 'properf_woo_metrics_snapshot', $result, HOUR_IN_SECONDS );
+
+		return $result;
 	}
 
 	/**
@@ -265,6 +274,8 @@ class ProPerf_Data_Collector {
 	 */
 	public function collect_and_push() {
 		require_once PROPERF_DIR . 'includes/class-bigquery-client.php';
+
+		delete_transient( 'properf_woo_metrics_snapshot' );
 
 		try {
 			$metrics = $this->get_data();
--- a/includes/class-data-collector.php
+++ b/includes/class-data-collector.php
@@ -14,6 +14,8 @@ if ( ! defined( 'ABSPATH' ) ) {
  */
 class ProPerf_Data_Collector {
 
+	const WOO_METRICS_CACHE_KEY = 'properf_woo_metrics_snapshot';
+
 	/**
 	 * Get collected data.
 	 *
@@ -96,7 +98,7 @@ class ProPerf_Data_Collector {
 		}
 
 		// Cached for 1 hour; invalidated on push and on relevant settings changes.
-		$cached = get_transient( 'properf_woo_metrics_snapshot' );
+		$cached = get_transient( self::WOO_METRICS_CACHE_KEY );
 		if ( false !== $cached ) {
 			return $cached;
 		}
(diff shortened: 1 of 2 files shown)


includes/class-data-collector.php
```diff
@@ -95,6 +95,12 @@ public function collect_woo_order_metrics() {
 			);
 		}
 
+		// Cached for 1 hour; invalidated on push and on relevant settings changes.
+		$cached = get_transient( 'properf_woo_metrics_snapshot' );
```
pokhiii: Small code-hygiene suggestion (non-blocking).

The string literal `'properf_woo_metrics_snapshot'` now appears in **5 places** across **2 files** (this file at lines 99, 221, 285, 298, and `class-admin-settings.php:53`). Two concerns:

**1. Single source of truth.** A class constant on `ProPerf_Data_Collector` would mean a future rename touches one line instead of five — and prevents the kind of silent-cache-break bug where someone updates the key in one spot and forgets the others:

```php
class ProPerf_Data_Collector {
    const WOO_METRICS_CACHE_KEY = 'properf_woo_metrics_snapshot';
    // ...
}
```

Then `get_transient( self::WOO_METRICS_CACHE_KEY )`, etc.

**2. Encapsulation.** Right now `ProPerf_Admin_Settings::bust_woo_metrics_cache()` at `class-admin-settings.php:53` hardcodes the cache key — but the cache is owned by the collector. The settings class is reaching into another class's implementation detail.

Cleaner: expose a public method on the collector and have settings call that:

```php
// in ProPerf_Data_Collector
public static function bust_metrics_cache() {
    delete_transient( self::WOO_METRICS_CACHE_KEY );
}

// in ProPerf_Admin_Settings
public static function bust_woo_metrics_cache() {
    ProPerf_Data_Collector::bust_metrics_cache();
}
```

Or, more directly, hook `ProPerf_Data_Collector::bust_metrics_cache` straight from `register_persistent_hooks()` — drop the settings-side method entirely.

Either way, the cache key stays in one class. Both improvements are small and orthogonal — feel free to split or batch.

includes/class-data-collector.php
```diff
@@ -257,6 +267,25 @@ public function record_qet_reading( $qet_ms ) {
...
 	}
 
+	/**
+	 * Refresh the baseline fields in the cached snapshot after a new QET reading is recorded.
+	 * Keeps the rest of the snapshot intact so the dashboard reflects the post-push baseline immediately.
+	 * Returns the updated baseline array, or null if no snapshot was cached.
+	 *
+	 * @return array|null {ms: int|null, source: string|null}, or null if no cached snapshot exists.
+	 */
+	private function refresh_cached_baseline() {
```
pokhiii, line 284: Nice addition — keeps the BigQuery row and the dashboard baseline aligned for the same push. Worth a brief docblock note on *when this actually changes anything*, so a future reader doesn't wonder if it's redundant.

My mental model: in steady state (locked baseline already exists), this is a no-op — `get_baseline_qet()` returns the locked value either way. The two real cases where it matters are:

1. **The new reading is the 10th post-archival reading** — it triggers the baseline lock-in inside `get_baseline_qet()`. Without `refresh_cached_baseline()`, the BigQuery row would still carry the `'post-archival-pending:10'` source from before the lock; with it, the row reflects the freshly-locked `'post-archival'` baseline.
2. **Lowest-10 fallback shifts** — if the new reading is lower than the previous 10th-lowest, the rolling baseline ticks down. Same alignment concern.

One extra sentence in the docblock would save the next person five minutes of "wait, doesn't `get_data()` already include the baseline?" thinking. Suggested:

```
 * Most calls are no-ops (locked baseline returns the same value either way).
 * Matters at lock-in (10th post-archival reading) and when the lowest-10
 * fallback shifts down — both move the baseline in ways the just-computed
 * snapshot won't reflect.
```

pokhiii (approved):
Closes the loop on the perf concern I'd raised back during PR #20. Clean implementation.

## ACs verified against #24

- ✅ Dashboard reads from cache (transient check at `class-data-collector.php:99-101`, 1h TTL via `HOUR_IN_SECONDS` at line 221)
- ✅ Cron and manual push always compute fresh (`collect_and_push` deletes the transient at line 298 before `get_data()`)
- ✅ No cache bleed into QET history (`record_qet_reading` is only called from `collect_and_push`, which bypasses cache)
- ✅ Dashboard display unchanged (same return shape preserved end-to-end)
- ✅ Non-WC metrics unaffected (only `collect_woo_order_metrics()` touched)

## Above-and-beyond improvements

Two things landed that aren't strictly in #24's scope but are correct calls:

1. **Cache invalidation covers both `update_option_*` AND `add_option_*` for both settings** — picks up first-ever save (when the option row doesn't exist yet) AND subsequent edits. Also covers both `properf_last_archival_date` AND `properf_archival_threshold_years`, since the threshold changes the `orders_older_than_threshold` count too. Not just archival date.

2. **`refresh_cached_baseline()` keeps the dashboard + BigQuery row consistent post-push.** Without it, the BigQuery row would reflect the pre-push baseline (computed before `record_qet_reading()` ran), while the dashboard's next live recompute would show the post-push baseline. Now they match. See inline note for a small docblock suggestion on this method.

## Cache invalidation completeness check

What busts the cache today:
- `update_option_/add_option_properf_last_archival_date`
- `update_option_/add_option_properf_archival_threshold_years`
- `collect_and_push()` deletes before recompute (cron + manual push)

What *doesn't* bust the cache:
- New orders / order deletions in WooCommerce — `total_orders`, `orders_older_than_threshold`, `latest_order_date` can be up to 1h stale on the dashboard. Per #24 this is the intended tradeoff ("Cache TTL: 1 hour"), so it's fine.
- A plugin version upgrade where the snapshot shape changes — would serve stale-shape data for up to 1h. TTL self-heals. Not worth handling.

## Race-condition awareness (minor, not blocking)

Concurrent dashboard view + manual push: both compute, last-write-wins on `set_transient`. In the worst case the dashboard's late `set_transient` overwrites the push's snapshot, so the dashboard could show stale values until next cache invalidation. Given cron is daily and manual pushes are rare, the probability is very low. Worth a one-line docblock note next to the cache write if you want belt-and-suspenders awareness for future maintainers — same pattern you already used on `record_qet_reading()`.

## Verdict

Approving. Two small inline suggestions for code hygiene — neither blocks the merge.

includes/class-data-collector.php
Ajay-Singh-Adhikari: Done. Added the explanation to the docblock as suggested.

includes/class-data-collector.php
Ajay-Singh-Adhikari: Done — extracted `WOO_METRICS_CACHE_KEY` as a class constant on `ProPerf_Data_Collector`. All five string literal usages replaced with `self::WOO_METRICS_CACHE_KEY`. Added `bust_metrics_cache()` as a public static on the collector and updated `register_persistent_hooks()` to hook it directly, removing the settings-side wrapper.

PR #33, opened 2026-07-14
Add DB size alert threshold with auto-suggestion and archival signal banner

### LINKED ISSUE #27: Archival readiness signal: order_item_meta size threshold with auto-suggestion
## Context

Phase 1 is complete — the plugin collects order metrics (order_item_meta size, orders older than retention threshold, QET, baseline QET) and pushes them to BigQuery. Data is flowing on MFM production as of June 9, 2026.

Phase 2 is making the plugin proactively signal archival readiness, before an engineer has to ask and before a user feels degradation.

---

## Problem with the current flow

An engineer who wants to know if archival is needed today has to:

1. Open the dashboard
2. Look at Current QET and Baseline QET
3. Decide manually if the gap is large enough to act on
4. Cross-check with the orders-older-than-threshold count
5. Make a judgment call

This is fully reactive. Degradation goes unnoticed until it becomes a support issue. There is no standardised judgment — two engineers looking at the same numbers reach different conclusions. Nobody watches the dashboard between engagements.

Goal: the tool should say &#34;a month from today your user will say the site is running slow — act now.&#34; Not after the user already said it.

---

## Why QET is not the right primary signal

QET was the first instinct because it directly measures DB query speed. The problem is it measures too many things at once.

1. **Traffic dependency** — The same DB with the same data runs a different QET at midnight versus peak hours. The cron fires at a fixed time. A high reading could be performance degradation or a busy server.

2. **Baseline timing problem** — The baseline locks after the first 10 readings. For MFM, ProPerf was deployed 17 months after the January 2025 archival. The baseline was set on a DB already at 12,500 MB — not the clean post-archival state. &#34;2× baseline&#34; fires at 2× an already-elevated starting point.

3. **No calibration reference** — MFM was catastrophic at 17.3 GB with 6.26s TTFB. But we don&#39;t know at what QET value that degradation began. 17.3 GB is &#34;when someone finally ran an audit,&#34; not &#34;when degradation started.&#34;

4. **Confirmed in production** — QET fluctuates with traffic volume, not just data volume. The automatic QET trigger was already dropped for this reason.

---

### Why these two metrics

**`order_item_meta` size (MB)**
- Traffic-independent — doesn&#39;t fluctuate day to day
- Changes slowly and consistently based on data volume
- Directly measures what archival removes
- Deterministic: 12,500 MB today will be 12,500 MB tomorrow regardless of traffic

**Orders older than retention threshold (count)**
- Confirms archivable data exists and is contributing to the size
- Without this condition, the table could be large from recent orders only — archiving wouldn&#39;t help
- Prevents false alerts: size must be high AND the data causing it must be archivable

**Together:** *&#34;The DB is getting bloated AND there is archivable data that would reduce it.&#34;*

Both conditions must be true simultaneously for the archival signal to fire.

---

## Auto-suggested threshold

The engineer needs to set the `order_item_meta` size threshold once per client. To calibrate this, the plugin estimates what the DB would look like after archival using data it already collects:

```
avg MB per order    = order_item_meta size ÷ total order count
archivable size     = avg MB per order × orders older than threshold
post-archival est.  = current size − archivable size
suggested threshold = post-archival estimate × 2  (100% headroom before alert fires, to reduce alert fatigue during initial calibration)
```

**Why × 2 and not × 1.5:** MFM's real numbers (18,400 MB post-archival estimate) match Kuldeep's planning signal of 18–20 GB at × 2. At × 1.5 the threshold would be ~13,280 MB — too early, causing alert fatigue before archival is actually needed. 100% headroom gives the engineer confidence the alert is meaningful when it fires.

**Example (MFM, hypothetical total_orders = 50,000):**

| Step | Calculation | Result |
|---|---|---|
| Avg MB/order | 12,500 ÷ 50,000 | 0.25 MB/order |
| Archivable size | 0.25 × 14,584 | ~3,646 MB |
| Post-archival estimate | 12,500 − 3,646 | ~8,854 MB |
| Suggested threshold | 8,854 × 2 | ~17,708 MB |

The settings field shows this derivation and pre-fills the value. The engineer can accept or override.

A **500 MB minimum floor** suppresses the suggestion for sites too small for archival planning — an informational message is shown instead.

---

## UX

**Experienced engineer view (settings page):**

```
Current order_item_meta: 12,500 MB
Total orders: 50,000 | Orders older than 2 years: 14,584
Estimated post-archival size: ~8,854 MB
Suggested threshold: 17,708 MB  ← editable
```

They understand the uniform MB/order assumption and can judge if it holds. They might set it lower for an earlier warning on a high-traffic site. They use the dashboard badge as confirmation, then look at raw numbers before recommending archival.

**Newbie view:**

Settings shows a pre-filled number with one line: *&#34;Alert when order_item_meta exceeds 17,708 MB — estimated based on current DB state.&#34;* They hit Save.

From then on, the dashboard shows:

| Badge | Meaning | Action |
|---|---|---|
| ✅ DB health: OK | Both conditions not met | Nothing to do |
| ⚠️ Archival recommended | Size > threshold AND archivable orders exist | Escalate or follow archival runbook |

The newbie&#39;s job is to notice the badge and act. The tool does the judgment.

---

## Implementation scope

1. **Add `total_order_count` metric** — one `COUNT(*)` query in `collect_woo_order_metrics()`. Needed for the avg MB/order calculation.

2. **Add `order_itemmeta_alert_threshold_mb` settings field** in `class-admin-settings.php` alongside existing retention window and archival date fields.

3. **Auto-suggestion computation** in the settings field renderer — reads current metrics, runs the formula, pre-fills the input.

4. **Two-condition evaluation** in `class-data-collector.php` — compare stored size against threshold, check orders older than threshold > 0.

5. **Push `archival_signal_active` (boolean) + `order_itemmeta_alert_threshold_mb` to BigQuery** in `format_for_bigquery()`.

---

## Acceptance criteria

- [ ] `total_order_count` is collected and pushed to BigQuery on every run
- [ ] `order_itemmeta_alert_threshold_mb` settings field exists in the WooCommerce Settings section, optional, accepts positive integer
- [ ] Field validates input — rejects zero, negative, or non-numeric values with an error; preserves previous value
- [ ] Settings page shows auto-suggested threshold with the derivation (current size, total orders, archivable orders, post-archival estimate)
- [ ] Sites with DB below 500 MB show "too small for archival planning" instead of a suggested value — 500 MB floor suppresses suggestions for sites too small for archival planning
- [ ] Plugin evaluates: if `order_itemmeta_size_mb > order_itemmeta_alert_threshold_mb` AND `orders_older_than_threshold > 0`, archival signal is active
- [ ] If threshold is not configured, no evaluation happens — dashboard shows raw numbers only
- [ ] `archival_signal_active` (boolean) and `order_itemmeta_alert_threshold_mb` are pushed to BigQuery alongside other metrics
- [ ] Dashboard badge shows ✅ DB health: OK or ⚠️ Archival recommended based on signal state

## Related

- #14 — archival signal dashboard (parent design issue, Phase 1 closed)
- #15 — QET collector
- #11 — server-side metrics tracking issue

### LINKED ISSUE #28: Surface archival readiness badge on dashboard when signal conditions are met
## Context

Issue #27 defines the two-condition archival signal: `order_item_meta size > configured threshold MB` AND `orders older than retention window > configured minimum count`. Both thresholds are set once per client in the plugin settings.

This issue is Step 3 — making that evaluation visible on the dashboard without requiring an engineer to interpret raw numbers.

## Problem

Even with order_item_meta size and orders older than threshold visible on the dashboard, an engineer still has to open the page, read the numbers, and decide. If nobody is watching — which is the normal state between engagements — degradation goes unnoticed until a performance complaint surfaces.

The plugin needs to surface the signal directly. The engineer should not have to derive the conclusion from raw numbers; the dashboard should state it.

### 1. Dashboard status badge

Add a clear archival readiness badge to the WooCommerce Order Metrics section:

| State | Badge |
|---|---|
| Threshold not configured | No badge — raw numbers only |
| Both conditions not met | ✅ **DB health: OK** |
| Size > threshold AND archivable orders > minimum count | ⚠️ **Archival recommended** |

The badge must be immediately visible without scrolling — the engineer opens the dashboard and sees the status at a glance.

### 2. WP Admin notice (optional, lower priority)

When the archival signal is active, show a persistent WP Admin notice on ProPerf pages so the signal is visible even without opening the full dashboard.

## What the signal means and does NOT mean

The alert means: the order_item_meta table has grown past the configured size threshold AND there are orders old enough to be archival candidates. Together, these indicate that archiving old orders is likely to reduce the table size meaningfully.

The signal does **not** trigger archival automatically. Archival modifies production data and requires human sign-off. The plugin's job is to make the recommendation obvious — the engineer takes action.

## Dependency

This issue depends on #27 (configurable order_item_meta threshold + minimum archivable orders count). The badge logic cannot be implemented until both threshold fields exist and `archival_signal_active` is being evaluated and pushed to BigQuery.

## Acceptance criteria

- [ ] Dashboard shows an archival readiness badge in the WooCommerce Order Metrics section
- [ ] Badge is absent when no threshold is configured
- [ ] Badge shows "DB health: OK" (green) when both conditions are not met
- [ ] Badge shows "Archival recommended" (amber) when `order_itemmeta_size_mb > order_itemmeta_alert_threshold_mb` AND `orders_older_than_threshold > min_archivable_orders_count`
- [ ] Badge evaluates against the cached snapshot (not a fresh live query) so it adds no DB load on dashboard load
- [ ] Status is consistent with what BigQuery data shows — no discrepancy between dashboard and Looker Studio

## Related

- #27 — configurable order_item_meta threshold (dependency)
- #14 — archival signal dashboard design
- #24 — dashboard caching (badge must read from cache)

Ajay-Singh-Adhikari:
Closes #27
Closes #28

## What changed

Before this PR, there was no way for an engineer to know whether a client&#39;s database needed archival without opening the dashboard, reading raw numbers, and making a manual judgment call. This PR adds a configurable size threshold for the `wp_woocommerce_order_itemmeta` table, evaluates an archival signal against it on every push, and surfaces the result as a dashboard banner — replacing the manual judgment with a clear automated signal.

## Settings: DB Size Alert Threshold

**Before:** No threshold setting existed. Archival readiness was left entirely to the engineer.

**After:** A `DB Size Alert Threshold` field is added under WooCommerce Settings with an MB/GB unit dropdown. The value is always stored in MB internally; the unit dropdown is submitted as a named form field and conversion from GB to MB happens server-side in the sanitize callback — JS is display-only.

When the field is empty, the settings page collects a live metrics snapshot and auto-suggests a threshold using this formula:

```
avg MB/order        = order_itemmeta_size ÷ total_orders
archivable size     = avg × orders_older_than_retention_window
post-archival (z)   = current_size − archivable_size
suggested threshold = z × 2
```

A 500 MB minimum floor suppresses the suggestion for sites too small for archival planning — an informational message is shown instead. If no orders are older than the retention window, the suggestion is skipped and the engineer is prompted to set it manually.

## Archival signal evaluation

**Before:** `order_itemmeta_size_mb` and `orders_older_than_threshold` were collected but never compared — no signal was evaluated.

**After:** On every metrics collection, the signal fires when both conditions are true simultaneously:
1. `order_itemmeta_size_mb >= alert_threshold_mb`
2. `orders_older_than_threshold > 0`

The second condition prevents false alerts when a large table is caused entirely by recent orders that cannot be archived yet. If no threshold is configured, no evaluation runs.

Result is pushed to BigQuery as:
- `woo_archival_signal_active` (BOOLEAN)
- `woo_alert_threshold_mb` (INTEGER)

## Dashboard signal banner

**Before:** No signal was visible on the dashboard. The engineer had to read raw numbers to make a call.

**After:** A banner appears above the metrics tables — immediately visible without scrolling. Three states:
- Hidden when no threshold is configured
- **DB health: OK** (green) — conditions are not met
- **Archival recommended** (red) — both conditions are met

The threshold is read live from the database on every dashboard load (not from the cached snapshot), so the banner reflects the current state immediately when the threshold is changed or cleared.

## Files changed

| File | Change |
|---|---|
| `class-admin-settings.php` | New `DB Size Alert Threshold` field, auto-suggestion logic, sanitize callback, cache-busting hooks |
| `class-admin-dashboard.php` | Archival signal banner rendered above metrics tables |
| `class-data-collector.php` | Signal evaluation in `collect_woo_order_metrics()`, three new fields in `format_for_bigquery()` |
| `class-bigquery-client.php` | Three new schema fields: `woo_total_orders`, `woo_archival_signal_active`, `woo_alert_threshold_mb` |
| `class-live-data.php` | Fallback defaults for the two new fields |
| `admin.css` | BEM classes for threshold input wrap and signal banner |
| `properf-wordpress-adapter.php` | Version bumped: 1.0.3 → 1.1.0 |

## Acceptance criteria

- [x] `DB Size Alert Threshold` field exists under WooCommerce Settings with MB/GB selector
- [x] Field validates input — rejects zero and negative values, preserves previous value on invalid input
- [x] Auto-suggestion pre-fills threshold from current DB snapshot using the formula when field is empty
- [x] Sites with DB below 500 MB show "too small for archival planning" instead of a low suggested value
- [x] Sites with no archivable orders show "set manually" message
- [x] Signal evaluates: `size >= threshold` AND `orders_older_than_threshold > 0`
- [x] If no threshold is configured, no evaluation runs
- [x] `woo_total_orders`, `woo_archival_signal_active`, and `woo_alert_threshold_mb` pushed to BigQuery on every run
- [x] Dashboard banner is visible above metrics tables
- [x] Banner is hidden when no threshold is configured
- [x] Banner reflects the current threshold state immediately — no stale cache

## Self-review fixes

A deeper review pass after the initial implementation caught and fixed 11 additional issues:

| # | File | Fix |
|---|---|---|
| 1 | `class-admin-dashboard.php` | `(float)` cast before `number_format` — PHP 8.1+ `TypeError` crash when stale transient has `null` for `order_itemmeta_size_mb` |
| 2 | `class-data-collector.php` | Added `archival_signal_active: false` and `alert_threshold_mb: null` to no-WC early return — prevented undefined-index notices in `format_for_bigquery()` on non-WC sites |
| 3 | `class-admin-settings.php` | Changed MB→GB display from `Math.round(x*100)/100` to `toFixed(4)` — prevents lossless round-trip drift (e.g. 1500 MB → 1.46 GB → 1495 MB on save) |
| 4 | `class-admin-settings.php` | Fixed `elseif ($suggested > 0)` → `elseif ($healthy > 0)` — wrong branch fired for very small DB sizes where `round($healthy * 2) = 0` |
| 5 | `class-data-collector.php` | Changed falsy `$itemmeta_size ?` check to `null !== $itemmeta_size` — DB permission errors returning `null` no longer silently set size to `0.0` and permanently suppress the archival signal |
| 6 | `class-admin-settings.php` | Transient write-back after cold collection — Settings page was running full DB queries on every load when transient expired, result was never cached |
| 7 | `class-admin-dashboard.php` | Dashboard now reads `archival_signal_active` and `alert_threshold_mb` from the woo metrics snapshot instead of recomputing live — removes duplicate derivation logic and single-sources the signal |
| 8 | `class-admin-settings.php` | Accurate error message on first-ever invalid save — previously said "Previous value restored" even when no prior value existed |
| 9 | `class-bigquery-client.php` | Added `ALLOW_FIELD_ADDITION` to load job — without it, any existing table with the old schema would fail on the next push |
| 10 | `class-admin-settings.php` | `step` changed from `0.01` to `any` for GB mode — browser was rejecting form submission when `toFixed(4)` produced 4-decimal values like `1.4648` |
| 11 | `class-bigquery-client.php` | Marked `woo_order_items_size_mb`, `woo_order_itemmeta_size_mb`, `woo_orders_older_than_threshold`, and `woo_query_execution_ms` as `NULLABLE` — all four receive `null` from the no-WC early return, causing load job failures on non-WC sites |

## Deploy note

This PR adds three new columns to the BigQuery schema: `woo_total_orders`, `woo_archival_signal_active`, and `woo_alert_threshold_mb`. Any existing BigQuery table receiving pushes will fail on the next run after this PR merges unless these columns are added first. The load job now includes `ALLOW_FIELD_ADDITION` as a safety net, but the DDL must still run first for tables where these columns don&#39;t exist yet.

**Ajay** is responsible for adding the following fields to the MFM BigQuery table schema before the next scheduled push. In the BigQuery console, open the table → Edit Schema → New fields → Edit as text, paste the JSON below, and save:

```json
[
  {"name": "woo_total_orders", "type": "INTEGER", "mode": "NULLABLE"},
  {"name": "woo_archival_signal_active", "type": "BOOLEAN", "mode": "NULLABLE"},
  {"name": "woo_alert_threshold_mb", "type": "INTEGER", "mode": "NULLABLE"}
]
```


Ajay-Singh-Adhikari:
## Spec redesigned — QET threshold replaced with order_item_meta size threshold

The original spec proposed a configurable QET alert threshold (ms). After reviewing production data from MFM, that approach was dropped in favour of `order_item_meta` size + orders older than threshold. Reasons:

**Why QET was dropped as the primary signal:**
- QET is traffic-dependent — the same DB produces different readings at midnight vs peak hours. The cron fires at a fixed time, so a high reading could mean degradation or just a busy server.
- The baseline locks after the first 10 readings. ProPerf was deployed 17 months after MFM's January 2025 archival, so the baseline was set on a DB already at 12,500 MB — not the clean post-archival state. "2× baseline" fires at 2× an already-elevated starting point.
- We have no reference for *when* degradation began on MFM. 17.3 GB is "when someone ran an audit," not "when performance started dropping." There's no calibration anchor.

**Why order_item_meta size + orders older than threshold:**
- Size is traffic-independent — 12,500 MB today is 12,500 MB tomorrow regardless of load
- The two-condition check prevents false alerts from recent order growth where archiving wouldn't help

**Signal condition:**
`order_item_meta size > [configurable size threshold MB]` AND `orders older than retention window > [configurable minimum count]`

**How the size threshold is configured:**

The settings page shows a breakdown above the threshold field — current table size, total orders, orders older than the retention window, and an estimated post-archival size. The threshold field is pre-filled using that estimate with 50% headroom added.

For an experienced engineer, the breakdown is a verification step. They read the numbers, judge whether the uniform MB/order assumption holds for that client, and override the value if needed — for instance setting a lower threshold for an earlier warning on a high-traffic site.

For an engineer who is not yet familiar with the client, the pre-filled value is the answer. They read the one-line description next to the field, confirm the number looks reasonable, and save. The plugin handles the judgment from that point — the dashboard shows either a green OK badge or an amber "Archival recommended" badge. No need to interpret raw numbers or compare QET values.

Ajay-Singh-Adhikari:
## Formula validation

**MFM:**
- 12,250 MB current, 411,614 total orders, ~105,766 older than threshold
- Formula gives ~18,400 MB threshold
- Kuldeep confirmed 18–20 GB is MFM's real archival planning signal ✓

**Sankalptaru:**
- 9.55 MB itemmeta, 7,688 total orders, 2,677 older than 8 years
- Formula gives ~12 MB — correctly suppressed by 500 MB floor
- Message shown: "DB is too small for archival planning"

---

## Resetting the DB size alert threshold

The threshold is calculated once from the current DB snapshot and stored as a fixed value. It does not recompute automatically — this is intentional so the reference point doesn't shift after archival.

There is no UI reset. To clear the threshold (e.g. to re-run the suggestion on a fresh DB state after archival), delete the option directly from the database:

```sql
DELETE FROM <prefix>_options WHERE option_name = 'properf_order_itemmeta_db_alert_threshold';
```

**Sankalptaru example:**
```bash
mysql -u root uat_sankalptaru_web -e "DELETE FROM staru_options WHERE option_name = 'properf_order_itemmeta_db_alert_threshold';"
```

After deleting, reload ProPerf > Settings — the field will be empty and the auto-suggestion will run again from the current snapshot. For sites where the DB is below 500 MB, no suggestion is shown and the threshold must be set manually.

---

includes/class-data-collector.php
```diff
@@ -414,18 +423,20 @@ public function format_for_bigquery( $metrics ) {
...
+			'woo_order_items_size_mb'         => $metrics['woo']['order_items_size_mb'],
+			'woo_order_itemmeta_size_mb'      => $metrics['woo']['order_itemmeta_size_mb'],
+			'woo_oldest_order_date'           => $metrics['woo']['oldest_order_date'],
+			'woo_latest_order_date'           => $metrics['woo']['latest_order_date'],
 			'woo_orders_older_than_threshold' => $metrics['woo']['orders_older_than_threshold'],
-			'woo_last_archival_date'     => $metrics['woo']['last_archival_date'],
-			'woo_query_execution_ms'     => $metrics['woo']['query_execution_ms'],
-			'woo_baseline_qet_ms'        => $metrics['woo']['baseline_qet_ms'],
+			'woo_last_archival_date'          => $metrics['woo']['last_archival_date'],
+			'woo_query_execution_ms'          => $metrics['woo']['query_execution_ms'],
```
pokhiii, line 439: **AC gap.** Issue #27's very first acceptance criterion is:

> `total_order_count` is collected and pushed to BigQuery on every run

Collected: yes — `$total_orders` is computed in `collect_woo_order_metrics()` and included in the returned `'woo' => [...]` array (visible at line 210). Pushed: no — this method doesn't map it into the BigQuery row, and `woo_total_orders` isn't in the schema (`class-bigquery-client.php:135-146`).

Suggested addition — one line here between the existing fields:

```php
'woo_orders_older_than_threshold' => $metrics['woo']['orders_older_than_threshold'],
'woo_total_orders'                => $metrics['woo']['total_orders'],
'woo_last_archival_date'          => $metrics['woo']['last_archival_date'],
```

And add to the schema in `class-bigquery-client.php`:

```php
array( 'name' => 'woo_total_orders', 'type' => 'INTEGER', 'mode' => 'NULLABLE' ),
```

(NULLABLE because the WC-not-active early return in the collector returns `null` for this field.)

With `woo_orders_older_than_threshold` alone, the dashboard-side ratio ("31.5% of orders are archivable") that the PR description mentions can't actually be computed downstream — you need both numerator and denominator in BigQuery. So this isn't just a spec-completeness thing; it's also what makes the archival-percentage analysis possible from the warehouse side.

**Deploy note also needs updating.** Any BigQuery table already receiving pushes will need `woo_total_orders` added via DDL before the next push after this PR merges. The current PR description doesn't call this out.

includes/class-admin-settings.php
```diff
@@ -359,6 +411,108 @@ public static function last_archival_date_field() {
...
+				$total       = (int) $woo['total_orders'];
+				$older       = (int) $woo['orders_older_than_threshold'];
+				$avg_mb      = $current_mb / $total;
+				$archivable  = $avg_mb * $older;
+				$healthy     = $current_mb - $archivable;
+				$suggested   = (int) round( $healthy * 2 );
+
+				if ( $suggested >= self::MIN_ARCHIVAL_THRESHOLD_MB ) {
+					$stored_mb         = $suggested;
+					$suggestion_notice = __( 'Pre-filled based on current DB state: estimated post-archival size × 2. You can adjust this value.', 'properf' );
```
pokhiii: **AC gap on the suggestion UX.**

Issue #27 is specific about *what* the suggestion notice should show. The UX example:

```
Current order_item_meta: 12,500 MB
Total orders: 50,000 | Orders older than 2 years: 14,584
Estimated post-archival size: ~8,854 MB
Suggested threshold: 13,280 MB  ← editable
```

The purpose (called out in the issue): *"They understand the uniform MB/order assumption and can judge if it holds."*

Current implementation shows just: *"Pre-filled based on current DB state: estimated post-archival size × 2."* — one sentence, no numbers. An engineer can't tell whether the suggestion is trustworthy without opening the dashboard, cross-referencing, doing the math themselves. That defeats the point of the auto-suggestion.

All four values are already computed above (`$current_mb`, `$total`, `$older`, `$healthy`). It's just a matter of surfacing them. Sketch:

```php
$suggestion_notice = sprintf(
    /* translators: 1: current size MB, 2: total orders, 3: archivable orders, 4: post-archival estimate MB */
    __( 'Pre-filled from current DB state: %1$s MB order_itemmeta across %2$s total orders (%3$s older than retention). Estimated post-archival size: ~%4$s MB. Suggested threshold = post-archival × 2.', 'properf' ),
    number_format( $current_mb, 2 ),
    number_format( $total ),
    number_format( $older ),
    number_format( $healthy )
);
```

Wording can be trimmed but the four numbers should be visible.

pokhiii (commented):
Substantial feature — closes #27 (archival readiness signal). The mechanics are right: two-condition evaluation, cache-busting on the new setting (add/update/delete), NULLABLE schema fields, sanitize callback with user feedback, 500 MB floor as a thoughtful UX addition. Version bump `1.0.3 → 1.1.0` correct for a new feature.

However, walking issue #27's AC list against the diff surfaces three real gaps and one robustness concern. Details inline; summary of what's covered here:

## AC verification against #27

- [x] Field exists in WooCommerce Settings, positive integer, sanitize callback, error on invalid ✓
- [x] 500 MB floor for suggestions ✓ (bonus, not in issue — worth adding to the issue if we want future PRs to preserve it)
- [x] No archivable orders → "set manually" message ✓
- [x] `woo_archival_signal_active` and `woo_alert_threshold_mb` pushed to BigQuery ✓
- [x] Dashboard banner with two states, hidden when no threshold ✓
- [x] Banner reflects current threshold state immediately ✓ (via live threshold read + cache-bust hooks)
- [ ] **`total_order_count` (`woo_total_orders`) pushed to BigQuery on every run** — AC #1 of #27. Collected in `collect_woo_order_metrics()` but never added to `format_for_bigquery()` or the schema. See inline note.
- [ ] **Settings page shows auto-suggestion *with the derivation*** — #27's UX example is explicit about showing the four numbers (current size, total orders, archivable orders, post-archival estimate) so the engineer can judge whether the uniform MB/order assumption holds. Current notice just says "Pre-filled based on current DB state: estimated post-archival size × 2." See inline note.
- [ ] **Signal formula divergence** — #27 says `suggested threshold = post-archival estimate × 1.5 (50% headroom)`. Code uses `× 2`. See inline note.

## WP best-practices audit

Clean:
- No new SQL — pure feature layered on existing collector paths.
- All new output escaped (`esc_html`, `esc_attr`, `esc_html_e`).
- Capability gate + nonce inherited from Settings API.
- Sanitize callback validates positivity and preserves prior value on invalid input (matches the pattern set on `sanitize_threshold_years`).
- `add_option_` + `update_option_` + `delete_option_` hooks all wired for the new option — full coverage.
- Schema fields marked NULLABLE on BigQuery side.

Minor:
- `sprintf( __( '...%s...%s...' ), ... )` in the banner strings would be more translator-friendly with positional placeholders (`%1$s`, `%2$s`). WPCS-level nit only.
- Signal is computed twice — once in `collect_woo_order_metrics()` (stored in the snapshot, pushed to BQ) and once in `render_dashboard()` (recomputed live). Since the cache is busted on threshold change, the dashboard could just trust `$woo_metrics['archival_signal_active']` and drop the live re-computation. Non-blocking design nit.

## Robustness

One inline note on the JS-based GB→MB conversion — a silent misconfiguration path exists if JS is disabled or fails. Details on that inline.

## Positive callouts

- Cache-busting on all three lifecycle events (add / update / delete) — better than the pattern used for the earlier settings, and correct.
- 500 MB floor is a good judgment call; the alternative low-value suggestions would have been misleading.
- PR description is genuinely detailed — before/after per file, files-changed table, AC checklist.

## Suggested path forward

1. Add `woo_total_orders` to the schema and push. Small.
2. Expand the suggestion notice to show the derivation (roughly matching #27's UX example).
3. Decide `× 1.5` vs `× 2` — update either the issue or the code to match.
4. The GB→MB conversion is the only non-issue-AC concern that has a real user impact; see inline for one server-side approach.

Once (1)–(3) are addressed and (4) is either fixed or explicitly deferred, this is approve-ready.

includes/class-admin-settings.php
```diff
@@ -359,6 +411,108 @@ public static function last_archival_date_field() {
...
+			$woo = isset( $snapshot['woo'] ) ? $snapshot['woo'] : null;
+
+			if ( $woo && ! empty( $woo['total_orders'] ) && $woo['total_orders'] > 0 && ! empty( $woo['orders_older_than_threshold'] ) && $woo['orders_older_than_threshold'] > 0 ) {
+				$current_mb  = (float) $woo['order_itemmeta_size_mb'];
+				$total       = (int) $woo['total_orders'];
+				$older       = (int) $woo['orders_older_than_threshold'];
+				$avg_mb      = $current_mb / $total;
+				$archivable  = $avg_mb * $older;
+				$healthy     = $current_mb - $archivable;
+				$suggested   = (int) round( $healthy * 2 );
```
pokhiii, line 457: **Spec vs implementation drift.**

Issue #27's formula section says:

> `suggested threshold = post-archival estimate × 1.5  (50% headroom before alert fires)`

The code uses `× 2`. The PR description also says `× 2`. So the code and the description agree — but they diverge from the issue.

Same pattern I've flagged on earlier PRs: if a design decision moved between issue and implementation, the issue should be the source of truth for the *decided* value. Two options:

1. If `× 2` is the right call (100% headroom is more forgiving, fewer false alerts), update issue #27 with the reasoning — something like "changed from ×1.5 to ×2 to reduce alert-fatigue risk during initial calibration."
2. If `× 1.5` was the right call (the issue's rationale of "50% headroom" matches the MFM example's 13,280 MB output), change the code.

Either works — just need the issue and the code to point to the same number.

Same consideration applies to the `>= 500` floor: it's a good addition but not in the issue. Worth adding to #27's ACs so a future PR touching this area preserves it.

includes/class-admin-settings.php
```diff
@@ -359,6 +411,108 @@ public static function last_archival_date_field() {
...
+				}
+				setUnit( select.value );
+				select.dataset.prev = select.value;
+			} );
+			form.addEventListener( 'submit', function () {
+				if ( select.value === 'gb' && input.value !== '' ) {
+					input.value = Math.round( parseFloat( input.value ) * 1024 );
+				}
+			} );
+		}());
```
pokhiii, line 531: **Robustness concern: silent misconfiguration if JS fails.**

The unit conversion happens entirely client-side. On form submit:

```js
form.addEventListener( 'submit', function () {
    if ( select.value === 'gb' && input.value !== '' ) {
        input.value = Math.round( parseFloat( input.value ) * 1024 );
    }
} );
```

Failure scenarios:
- Ad blocker / security tool strips inline scripts (some enterprise setups do this)
- JS error earlier on the page prevents this handler from running
- Engineer submits with unit dropdown showing GB, but the JS never converted

In any of those, an engineer setting "5 GB" would submit `5`, and `sanitize_alert_threshold_mb()` sees `5`, validates positivity (`5 > 0` — pass), stores 5 MB. Signal fires immediately because every store's `order_itemmeta` is well over 5 MB. **Silent misconfiguration**: the engineer thinks they set a 5 GB threshold; they actually set 5 MB.

Suggested server-side approach — pass the unit as a real form field, convert in the sanitize callback:

```php
// In the renderer: replace the plain <select> with a real named field
<select id="properf_db_alert_threshold_unit" name="properf_order_itemmeta_db_alert_threshold_unit">
    <option value="mb">MB</option>
    <option value="gb">GB</option>
</select>

// In sanitize_alert_threshold_mb
public static function sanitize_alert_threshold_mb( $value ) {
    if ( '' === $value || null === $value ) {
        return '';
    }
    $unit = isset( $_POST['properf_order_itemmeta_db_alert_threshold_unit'] )
        ? sanitize_key( wp_unslash( $_POST['properf_order_itemmeta_db_alert_threshold_unit'] ) )
        : 'mb';
    $numeric = floatval( $value );
    if ( 'gb' === $unit ) {
        $numeric *= 1024;
    }
    $mb = (int) round( $numeric );
    if ( $mb <= 0 ) {
        // ... existing error branch ...
    }
    return $mb;
}
```

Then the JS becomes progressive-enhancement only (converts the *displayed* value for readability), not the source of truth for the submission.

Current JS-based approach also has a smaller wart: it uses `Math.round` for both directions of conversion, so a user entering `2.5 GB` (2560 MB) → toggling to MB → back to GB may end up at `2.5` still, but user entering `2.51 GB` → MB (2570) → GB (2.51) survives round-trips only by the `Math.round(x*100)/100` safety. Not broken, just fiddly. Server-side conversion sidesteps the whole thing.

includes/class-data-collector.php
Ajay-Singh-Adhikari: **Fixed:** Added `woo_total_orders` to `format_for_bigquery()` in `class-data-collector.php` and to the inline schema in `class-bigquery-client.php`. 
- Also updated the PR description and deploy note to include all three new columns. Verified in BigQuery

includes/class-admin-settings.php
Ajay-Singh-Adhikari: **Fixed:** The suggestion notice now surfaces all four values using `sprintf` with positional placeholders — current size MB, total orders, orders older than retention, and post-archival estimate. 
- Verified locally.

includes/class-admin-settings.php
Ajay-Singh-Adhikari: Updated issue #27 with the reasoning for ×2.

includes/class-admin-settings.php
Ajay-Singh-Adhikari: **Fixed:** The unit is now a real named form field `(name="properf_order_itemmeta_db_alert_threshold_unit")`, and GB→MB conversion happens in `sanitize_alert_threshold_mb()` server-side. JS is now display-only. Also added an explicit allowlist check — any unit value outside ['mb', 'gb'] defaults to mb before conversion. 
- Verified locally.

includes/class-data-collector.php
pokhiii: Verified — `woo_total_orders` added at `includes/class-data-collector.php:437` (in `format_for_bigquery()`) and at `includes/class-bigquery-client.php:143` (schema, `INTEGER` + `NULLABLE`). PR description's "Files changed" table and "Deploy note" updated to include all three new columns with explicit DDL. Resolving from my side.

includes/class-admin-settings.php
pokhiii: Verified at `class-admin-settings.php:461-468`. `sprintf` with positional placeholders (`%1$s`–`%4$s`) surfaces all four numbers — current itemmeta MB, total orders, orders older than retention, post-archival estimate. Includes the `/* translators: */` comment above the string so translators know the placeholder order. Bonus catch: also fixed the `elseif ( $suggested > 0 )` → `elseif ( $healthy > 0 )` bug at line 469, which was routing very small DBs (where `round($healthy * 2) = 0`) to the wrong branch. Resolving from my side.

includes/class-admin-settings.php
pokhiii: Verified in issue #27's updated body — the `× 2` formula is now the documented value, with an explicit "Why × 2 and not × 1.5" block calling out the MFM data point (~18,400 MB post-archival matches Kuldeep's 18–20 GB planning signal at ×2, while ×1.5 would fire too early). MFM example rows updated to 17,708 MB. Newbie UX text updated. The 500 MB floor also made it into the issue's AC list — good.

Issue and code now point to the same number. Resolving from my side.

includes/class-admin-settings.php
pokhiii: Verified end-to-end. The unit is now a real named form field (`<select … name="properf_order_itemmeta_db_alert_threshold_unit">` at `class-admin-settings.php:489`), and `sanitize_alert_threshold_mb()` at line 386-418 does the conversion server-side:

- `sanitize_key( wp_unslash( $_POST[...] ) )` on the unit value.
- Explicit allowlist: `in_array( $unit, array( 'mb', 'gb' ), true )` with fallback to `'mb'`. This is the belt-and-suspenders check I like to see.
- `floatval( $value ) * 1024` for GB → MB.

The JS is now display-only progressive enhancement — if it fails, the form still submits a valid unit and the sanitizer converts correctly. Silent-misconfiguration path is closed. Resolving from my side.

pokhiii (approved):
All four inline items landed cleanly — see per-thread replies for verification.

What stood out this round is the self-review discipline. Ajay ran a full second pass after addressing my comments and caught 11 additional issues, several of which would have been real bugs in production:

- **`ALLOW_FIELD_ADDITION` on the BigQuery load job** (`class-bigquery-client.php:156`) — this closes an outstanding concern I'd flagged all the way back on PR #20 that had been carried as a manual-runbook step ever since. Now BigQuery auto-adds new columns and the load job survives future schema additions. Big improvement.
- **NULLABLE mode on `woo_order_items_size_mb`, `woo_order_itemmeta_size_mb`, `woo_orders_older_than_threshold`, `woo_query_execution_ms`** — all four receive `null` from the no-WC early return, so without NULLABLE the load job would have failed on any non-WC site. Would only have surfaced when someone deployed this to a WC-less install.
- **`(float)` cast before `number_format`** on the dashboard banner — prevents PHP 8.1 `TypeError` when a stale transient has `null` for `order_itemmeta_size_mb`.
- **`elseif ( $healthy > 0 )` correction** — was `$suggested > 0`, which routed very small DBs (where `round($healthy * 2) = 0`) to the wrong branch. Same fix also came up on thread 2.
- **Dashboard reads `archival_signal_active` from the snapshot** instead of recomputing live — closes the "duplicate derivation logic" design nit I flagged in the summary body last time.
- **`toFixed(4)` for MB→GB display + `step="any"` for GB mode** — prevents lossy round-trips like `1500 MB → 1.46 GB → 1495 MB on save`. Cheap fix, real bug for anyone editing an existing threshold.
- **`null !== $itemmeta_size` guard** instead of a falsy check — DB permission errors returning `null` no longer silently coerce to `0.0` and permanently suppress the archival signal.

That's the kind of pass I'd want done on every non-trivial PR — take a step back after addressing feedback and reread the whole change with fresh eyes. Genuinely good engineering discipline.

Approving.

Issue #35, opened 2026-07-20
Define alert types, included metrics, and notification frequency for the alert system


Ajay-Singh-Adhikari:
## Context

Issue #34 covers building the external notification system. Before implementation begins, the scope of the alert system needs to be clearly defined — what types of alerts exist, which metrics trigger them, and how often they fire. Without this, we risk building a system that either spams or under-alerts.

### 1. Alert types

| Alert type | Trigger | BigQuery | Status |
|---|---|---|---|
| Archival recommended — predictive | Projected growth trend crosses `order_itemmeta` threshold within ~4 weeks (based on `db_size_history`/QET history growth rate) | Yes | To be defined |
| Archival recommended — reactive | `order_itemmeta_size >= threshold` AND `orders_older_than_threshold > 0` AND current QET exceeds baseline by a configurable percentage | Yes | Defined in #34 (threshold-only version); QET condition to be added |
| Abnormal DB size growth | Week-over-week DB size growth exceeds N× the rolling average weekly growth | Yes — `total_db_size_mb` collected in #11 | Defined in #34 |
| Abnormal QET growth | Current week's average QET exceeds last week's average QET by a configurable percentage | Yes — `query_execution_ms` collected pre-#11 | To be defined |
| High hook count | Total registered hook callbacks exceed a configurable threshold (suggested: 8,000) | Yes — collected in #11 | Defined in #34 |
| High active plugin count | Active plugin count exceeds a configurable threshold (suggested: 40) | Yes — collected in #11 | Defined in #34 |
| High inactive plugin count | Inactive plugin count exceeds a configurable threshold | Yes — collected in #11 | To be defined |
| Unexpected DB table growth | A non-WooCommerce table exceeds a configurable size threshold — flags data bloat from logging plugins or custom code, not orders | Yes — collected in #11 | To be defined |
| Plugin vulnerability detected | One or more active plugins have a known CVE (via WP.org vulnerability API) | No — alert only | Split out to #41 (wishlist, not yet approved) |

**Why the new alert types matter:**

These alerts exist because DB size alone doesn't tell you *why* a site is slow. Without them, every slow site gets the same response: check the orders, recommend archival. With them, the diagnosis becomes systematic:

[...]

Issue #34, opened 2026-07-20
Send external notification when archival signal fires


Ajay-Singh-Adhikari:
## Context

The archival signal is now evaluated and surfaced as a dashboard banner (#28, #33). However, the banner is only visible when someone actively logs into the WordPress admin and opens the ProPerf dashboard. In normal operations between engagements, nobody is watching — so the signal goes unnoticed until it's too late.

## Problem

The plugin needs to push a notification outside WordPress when the archival signal fires, so the right person is alerted proactively without having to log in and check.

## What to build

When `archival_signal_active` evaluates to `true` during a metrics collection run, send a notification to the configured platform. The notification should include:
- The site URL
- Current order itemmeta size (MB)
- Configured threshold (MB)
- Number of orders older than the retention window

The notification platform and recipient(s) are TBD — pending confirmation from MFM's team (see acceptance criteria).

## Dependency

Depends on #33 — `archival_signal_active` must be evaluated before this notification can be triggered.

## Acceptance criteria

- [x] Confirm with Kuldeep and Deepak which notification platform MFM already uses (Slack, email, SMS, or other)
- [x] Confirm who should receive the alert — specific person or team channel
- [x] Implement notification delivery to the confirmed platform when `archival_signal_active` is `true`
- [ ] Notification is not sent if no threshold is configured
- [x] Notification is not sent on every push if the signal is already active — only fires once when the signal transitions from inactive to active (no alert spam)
- [ ] Notification includes site URL, current itemmeta size, threshold, and archivable order count

## Related

- #28 — dashboard archival signal banner (done)
- #33 — signal evaluation and BigQuery push (done)

Ajay-Singh-Adhikari:
**Status: on hold, not closed.**

The team's primary DB-health monitoring workflow is Looker Studio dashboards (DB size vs. archival threshold trend, a single-row Archival Status indicator) rather than the plugin's own admin dashboard or a push-notification system. Whether this email alert system is still needed depends on whether dashboard-checking happens on a reliable cadence or is ad-hoc — that call hasn't been made yet.

Keeping this open rather than closing it: the original gap this was meant to solve (nobody watching, DB grows unnoticed until performance drops) is still a real risk if dashboard checks aren't consistent. The alert-system code itself already exists on the `feat/configure_alert_system` branch, unmerged, not being actively worked on for now.

PR #40, opened 2026-07-20
Add server-side metrics: plugin count, hook count, DB table sizes

### LINKED ISSUE #11: Add server-side WordPress metrics — plugin count, hook count, DB sizes, order table health

Ajay-Singh-Adhikari:
Closes #11

## What changed

Before this PR, the plugin collected autoloaded options and WooCommerce order metrics — enough to know the DB was large, but not enough to know *why* a site was slow. Every slow site looked the same and the engineer had to guess whether the problem was data, code, or plugin bloat. This PR adds three server-level metrics — plugin count, hook count, and DB table sizes — that turn guessing into a systematic triage.

### Server metrics collection

`collect_server_metrics()` in `ProPerf_Data_Collector` now merges two independently-collected, independently-cached pieces:

- **`collect_plugin_stats()`** — active + inactive plugin count (via `array_intersect()` against `get_plugins()` and the `active_plugins` option, including network-activated plugins on multisite) and total registered hook callbacks (all callbacks across every `WP_Hook` entry in `$wp_filter`). Cached under `PLUGIN_METRICS_CACHE_KEY`, busted on plugin activate/deactivate/install/delete.
- **`collect_table_stats()`** — per-table DB sizes from `information_schema.TABLES`, ordered largest first, plus the total DB size and any table-name encoding warning. Table names containing non-UTF-8 characters are sanitized via `iconv` (guarded — hosts without ext-iconv fall back to the raw name instead of fataling) before display. Cached under `SERVER_METRICS_CACHE_KEY`, on its own TTL, not tied to plugin events.

These two are cached separately because they have different invalidation triggers: plugin toggles shouldn't force an expensive `information_schema` re-scan, and DB table growth (general writes) has nothing to do with plugin state.

`get_data()` merges all three collectors. `format_for_bigquery()` includes 4 new scalar columns pushed on every sync — per-table sizes are dashboard-only.

### Dashboard

Two new sections added to **ProPerf → Dashboard**:
- **Server Metrics** — active/inactive plugin count, hook count, total DB size
- **Top 10 Database Tables by Size** — largest tables only, capped at 10 rows to keep it readable

Per-table sizes are not pushed to BigQuery.

### BigQuery schema migration

**Responsible: Ajay**

The load job is configured with `ALLOW_FIELD_ADDITION`, so these 4 new columns should be added automatically on the first push after the plugin is updated. If they aren't, add them manually in BigQuery Console: open the table → **Edit Schema → Edit as text**, add the following entries and save:

```json
{ "name": "active_plugin_count", "type": "INTEGER", "mode": "NULLABLE" },
{ "name": "inactive_plugin_count", "type": "INTEGER", "mode": "NULLABLE" },
{ "name": "hook_count", "type": "INTEGER", "mode": "NULLABLE" },
{ "name": "total_db_size_mb", "type": "FLOAT", "mode": "NULLABLE" }
```

## Code quality / correctness fixes

- **Stale transient post-deploy** — cache is now busted on plugin activation via `register_activation_hook` so the dashboard reflects fresh data immediately after update.
- **`activated_plugin`/`deactivated_plugin`/`deleted_plugin`/plugin-install hooks all wired** — previously only activate/deactivate busted the plugin cache; installing a new (inactive) plugin or deleting one also changes the plugin count but wasn't covered. Added `deleted_plugin` and an `upgrader_process_complete` install-type check alongside the existing activate/deactivate hooks, and moved all of them off `admin_init` so WP-CLI and REST paths are covered too.
- **`collect_and_push()` now busts all caches via `bust_metrics_cache()`** — previously listed transients individually and could drift out of sync with the actual set of caches; now calls the single aggregator.
- **`iconv()` unguarded → uncaught fatal** — `iconv()` was called with no `function_exists()` check, unlike the `get_plugins()` guard two lines above it. On a PHP build without ext-iconv this threw an `Error` (not an `Exception`), which the surrounding `catch ( Exception $e )` in both the dashboard render and the cron push does not catch — the whole page/push fatals instead of degrading. Now guarded the same way `get_plugins()` is.
- **Multisite undercounts active plugins** — `active_plugin_count` only read the per-site `active_plugins` option, missing network-activated plugins (`active_sitewide_plugins`), which were silently counted as inactive. Now included when `is_multisite()`.
- **`db_table_sizes` key collision silently dropped a table** — two distinct corrupted table names that both sanitize to the same string overwrote each other in the array while `total_db_size_mb` still counted both, so the displayed rows no longer summed to the displayed total. Now accumulates instead of overwriting on collision.
- **Fully-invalid table names were excluded from the total, not just the display** — a table name `iconv` couldn't clean at all (`false`/`''`) skipped the row before it was added to `total_db_size_mb`, silently undercounting real DB size. The size is now always added to the total; only the per-table display row is skipped when the name is unusable.
- **Encoding-warning option retired** — `properf_table_name_encoding_warning` was autoloaded (no `false` argument, unlike other options in this file), and required a get/update/delete sync dance on every cache read. It only had one consumer (the dashboard render, which already has fresh `$metrics` in hand), so it's gone entirely — the warning now lives at `$metrics['server']['table_name_warning']`, read directly, with no separate option and no stray top-level key leaking through `get_data()`'s `array_merge()`.
- **Settings field double-cached WooCommerce metrics** — `order_itemmeta_alert_threshold_mb_field()` checked the WOO transient itself, called `collect_woo_order_metrics()` (which checks and sets the same transient internally), then set it a third time with the identical value. Removed the redundant wrapper — the collector already owns its own caching.
- **`$wp_filter` PHP 8 fatal** — the hook callback counter accessed `->callbacks` on every `$wp_filter` entry without checking it's a `WP_Hook` object. Added `instanceof WP_Hook` guard.
- **`number_format(null)` crash on PHP 8.2** — `order_itemmeta_size_mb` can be null when the WooCommerce itemmeta table does not exist. The archival signal banner now requires a non-null size before rendering, and the table row shows `—` instead of crashing.
- **`$_POST` unit read in sanitize callback** — the alert threshold sanitizer reads the unit dropdown from `$_POST`, which is empty in WP-CLI and REST API contexts. Added `!empty($_POST)` guard so non-HTTP callers don't silently store the wrong threshold.
- **Misleading notice on fresh store** — the threshold field showed "No archivable orders found" on a zero-order store. It now shows "No orders yet" when `total_orders` is also zero.
- **`validate_active_plugins()` replaced with `array_intersect()`** — the original writes back to the DB and is unsafe in cron context. Replaced with a read-only `array_intersect()` against `array_keys(get_plugins())`.
- **Cache busting scoped** — split into `bust_woo_metrics_cache()`, `bust_autoload_metrics_cache()`, `bust_plugin_metrics_cache()`, and `bust_server_metrics_cache()` so each setting or event only invalidates the data it actually affects.
- **`collect_autoloaded_options()` uncached** — added 1-hour transient cache with `AUTOLOAD_METRICS_CACHE_KEY`, consistent with the other collectors.
- **Duplicate `properf_bq_last_sync` write** — removed from the BigQuery client; `collect_and_push()` already writes it.
- **Hardcoded cache key string in settings** — replaced raw string with `ProPerf_Data_Collector::WOO_METRICS_CACHE_KEY` constant.
- **`db_table_sizes` not pushed to BigQuery** — dropped from `format_for_bigquery()` and the BigQuery client's load schema.
- **Variable alignment** — fixed inconsistent spacing in `render_dashboard()` variable assignments; moved `$table_name_warning` assignment to the variables block before the template.

## Future work landing cleanly on this structure

- #35 — high hook count, high plugin count, and abnormal DB size growth alert types now have the data they need to fire from BigQuery (`hook_count`, `active_plugin_count`, `total_db_size_mb`)
- #34 — external notification delivery can build on top of these metrics once #35's thresholds are confirmed

## Acceptance criteria
- [x] **ProPerf → Dashboard** shows Server Metrics section with active/inactive plugin count, hook count, total DB size — verified on sankalptaru
- [x] **Top 10 Database Tables by Size** section shows at most 10 rows, ordered largest first — verified
- [x] Push to BigQuery succeeds and all 4 new scalar columns are non-null in the latest row — verified
- [ ] Dashboard renders without errors on a site where WooCommerce is not active
- [ ] Activating/updating the plugin clears the plugin metrics cache immediately; WooCommerce setting changes do not trigger an `information_schema` scan
- [ ] Setting the alert threshold via WP-CLI stores the correct MB value
- [ ] Dashboard encoding warning appears when a table name contains non-UTF-8 characters and clears when the condition is resolved
- [ ] Installing (without activating) or deleting a plugin updates the plugin count without waiting for the 1-hour cache TTL

### DIFF
--- a/includes/class-data-collector.php
+++ b/includes/class-data-collector.php
@@ -271,6 +271,10 @@ class ProPerf_Data_Collector {
 		$active_count   = count( $active_keys );
 		$inactive_count = count( $all_plugins ) - $active_count;
 
+		// Counts callbacks registered in $wp_filter at the moment this method runs
+		// (typically cron context or admin dashboard). Not "hooks per frontend
+		// page" — frontend-only hooks registered later in the request lifecycle
+		// aren't counted. Meant as a directional signal for plugin-stack heaviness.
 		global $wp_filter;
 		$hook_count = 0;
 		foreach ( $wp_filter as $hook ) {
@@ -412,6 +416,7 @@ class ProPerf_Data_Collector {
 
 		delete_transient( self::WOO_METRICS_CACHE_KEY );
 		delete_transient( self::SERVER_METRICS_CACHE_KEY );
+		delete_transient( self::AUTOLOAD_METRICS_CACHE_KEY );
 
 		try {
 			$metrics = $this->get_data();
@@ -506,6 +511,9 @@ class ProPerf_Data_Collector {
 
 	public static function bust_woo_metrics_cache() {
 		delete_transient( self::WOO_METRICS_CACHE_KEY );
+	}
+
+	public static function bust_autoload_metrics_cache() {
 		delete_transient( self::AUTOLOAD_METRICS_CACHE_KEY );
 	}
 
@@ -515,6 +523,7 @@ class ProPerf_Data_Collector {
 
 	public static function bust_metrics_cache() {
 		self::bust_woo_metrics_cache();
+		self::bust_autoload_metrics_cache();
 		self::bust_server_metrics_cache();
 	}
 
--- a/includes/class-data-collector.php
+++ b/includes/class-data-collector.php
@@ -493,12 +493,14 @@ class ProPerf_Data_Collector {
 	 * @return array Formatted data for BigQuery.
 	 */
 	public function format_for_bigquery( $metrics ) {
-		$site_url         = get_site_url();
-		$timestamp        = gmdate( 'Y-m-d H:i:s' );
-		$table_sizes_json = wp_json_encode( $metrics['server']['db_table_sizes'] );
-		if ( false === $table_sizes_json ) {
-			error_log( 'ProPerf Error: Failed to JSON-encode db_table_sizes for ' . $site_url . ' — possible non-UTF-8 table name. Sending empty object to BigQuery.' );
-			$table_sizes_json = '{}';
+		$site_url    = get_site_url();
+		$timestamp   = gmdate( 'Y-m-d H:i:s' );
+		$table_sizes = array();
+		foreach ( $metrics['server']['db_table_sizes'] as $name => $size_mb ) {
+			$table_sizes[] = array(
+				'name'    => $name,
+				'size_mb' => $size_mb,
+			);
 		}
 
 		return array(
@@ -521,7 +523,7 @@ class ProPerf_Data_Collector {
 			'inactive_plugin_count'           => $metrics['server']['inactive_plugin_count'],
 			'hook_count'                      => $metrics['server']['hook_count'],
 			'total_db_size_mb'                => $metrics['server']['total_db_size_mb'],
-			'db_table_sizes_json'             => $table_sizes_json,
+			'db_table_sizes'                  => $table_sizes,
 		);
 	}
 }
--- a/includes/class-data-collector.php
+++ b/includes/class-data-collector.php
@@ -252,10 +252,10 @@ class ProPerf_Data_Collector {
 		if ( ! function_exists( 'get_plugins' ) ) {
 			require_once ABSPATH . 'wp-admin/includes/plugin.php';
 		}
-		validate_active_plugins();
 		$all_plugins    = get_plugins();
 		$active_plugins = get_option( 'active_plugins', array() );
-		$active_count   = count( $active_plugins );
+		$active_keys    = array_intersect( $active_plugins, array_keys( $all_plugins ) );
+		$active_count   = count( $active_keys );
 		$inactive_count = count( $all_plugins ) - $active_count;
 
 		global $wp_filter;
--- a/includes/class-data-collector.php
+++ b/includes/class-data-collector.php
@@ -14,8 +14,9 @@ if ( ! defined( 'ABSPATH' ) ) {
  */
 class ProPerf_Data_Collector {
 
-	const WOO_METRICS_CACHE_KEY    = 'properf_woo_metrics_snapshot';
-	const SERVER_METRICS_CACHE_KEY = 'properf_server_metrics_snapshot';
+	const WOO_METRICS_CACHE_KEY       = 'properf_woo_metrics_snapshot';
+	const SERVER_METRICS_CACHE_KEY    = 'properf_server_metrics_snapshot';
+	const AUTOLOAD_METRICS_CACHE_KEY  = 'properf_autoload_metrics_snapshot';
 
 	/**
 	 * Get collected data.
@@ -36,6 +37,11 @@ class ProPerf_Data_Collector {
 	 * @return array Autoloaded options data.
 	 */
 	public function collect_autoloaded_options() {
+		$cached = get_transient( self::AUTOLOAD_METRICS_CACHE_KEY );
+		if ( false !== $cached ) {
+			return $cached;
+		}
+
 		global $wpdb;
 		$autoload_clause = "autoload IN ('yes', 'on', 'auto', 'auto-on')";
 
@@ -65,7 +71,7 @@ class ProPerf_Data_Collector {
 			}
 		}
 
-		return array(
+		$result = array(
 			'autoloaded_option' => array(
 				'count'         => $autoloaded_option_count,
 				'size_bytes'    => $autoloaded_option_size_bytes,
@@ -73,6 +79,8 @@ class ProPerf_Data_Collector {
 				'error'         => null,
 			),
 		);
+		set_transient( self::AUTOLOAD_METRICS_CACHE_KEY, $result, HOUR_IN_SECONDS );
+		return $result;
 	}
 
 	/**
@@ -505,6 +513,7 @@ class ProPerf_Data_Collector {
 	public static function bust_metrics_cache() {
 		delete_transient( self::WOO_METRICS_CACHE_KEY );
 		delete_transient( self::SERVER_METRICS_CACHE_KEY );
+		delete_transient( self::AUTOLOAD_METRICS_CACHE_KEY );
 	}
 
 	/**
--- a/includes/class-data-collector.php
+++ b/includes/class-data-collector.php
@@ -506,14 +506,18 @@ class ProPerf_Data_Collector {
 		);
 	}
 
-	/**
-	 * Delete the cached WooCommerce metrics snapshot.
-	 * Called by settings hooks when archival date or threshold changes.
-	 */
-	public static function bust_metrics_cache() {
+	public static function bust_woo_metrics_cache() {
 		delete_transient( self::WOO_METRICS_CACHE_KEY );
-		delete_transient( self::SERVER_METRICS_CACHE_KEY );
 		delete_transient( self::AUTOLOAD_METRICS_CACHE_KEY );
+	}
+
+	public static function bust_server_metrics_cache() {
+		delete_transient( self::SERVER_METRICS_CACHE_KEY );
+	}
+
+	public static function bust_metrics_cache() {
+		self::bust_woo_metrics_cache();
+		self::bust_server_metrics_cache();
 	}
 
 	/**
--- a/includes/class-data-collector.php
+++ b/includes/class-data-collector.php
@@ -246,6 +246,11 @@ class ProPerf_Data_Collector {
 	public function collect_server_metrics() {
 		$cached = get_transient( self::SERVER_METRICS_CACHE_KEY );
 		if ( false !== $cached ) {
+			if ( ! empty( $cached['_had_invalid_names'] ) ) {
+				update_option( 'properf_table_name_encoding_warning', $cached['_had_invalid_names'] );
+			} else {
+				delete_option( 'properf_table_name_encoding_warning' );
+			}
 			return $cached;
 		}
 
@@ -303,8 +308,10 @@ class ProPerf_Data_Collector {
 			delete_option( 'properf_table_name_encoding_warning' );
 		}
 
-		$result = array(
-			'server' => array(
+		$warning_timestamp = $had_invalid_names ? gmdate( 'Y-m-d H:i:s' ) : '';
+		$result            = array(
+			'_had_invalid_names' => $warning_timestamp,
+			'server'             => array(
 				'active_plugin_count'   => $active_count,
 				'inactive_plugin_count' => $inactive_count,
 				'hook_count'            => $hook_count,
@@ -312,6 +319,11 @@ class ProPerf_Data_Collector {
 				'total_db_size_mb'      => round( $total_db_size_mb, 4 ),
 			),
 		);
+		if ( $had_invalid_names ) {
+			update_option( 'properf_table_name_encoding_warning', $warning_timestamp );
+		} else {
+			delete_option( 'properf_table_name_encoding_warning' );
+		}
 		set_transient( self::SERVER_METRICS_CACHE_KEY, $result, HOUR_IN_SECONDS );
 		return $result;
 	}
--- a/includes/class-data-collector.php
+++ b/includes/class-data-collector.php
@@ -16,6 +16,7 @@ class ProPerf_Data_Collector {
 
 	const WOO_METRICS_CACHE_KEY       = 'properf_woo_metrics_snapshot';
 	const SERVER_METRICS_CACHE_KEY    = 'properf_server_metrics_snapshot';
+	const PLUGIN_METRICS_CACHE_KEY    = 'properf_plugin_metrics_snapshot';
 	const AUTOLOAD_METRICS_CACHE_KEY  = 'properf_autoload_metrics_snapshot';
 
 	/**
@@ -247,18 +248,15 @@ class ProPerf_Data_Collector {
 	}
 
 	/**
-	 * Collect server-level metrics: plugin counts, hook callback count, and DB table sizes.
+	 * Collect active/inactive plugin count and hook callback count.
+	 * Cached separately from table sizes: this data changes on plugin (de)activation,
+	 * not on general DB writes, so it has its own invalidation triggers and TTL.
 	 *
-	 * @return array Server metrics keyed under 'server'.
+	 * @return array {active_plugin_count, inactive_plugin_count, hook_count}
 	 */
-	public function collect_server_metrics() {
-		$cached = get_transient( self::SERVER_METRICS_CACHE_KEY );
+	private function collect_plugin_stats() {
+		$cached = get_transient( self::PLUGIN_METRICS_CACHE_KEY );
 		if ( false !== $cached ) {
-			if ( ! empty( $cached['_had_invalid_names'] ) ) {
-				update_option( 'properf_table_name_encoding_warning', $cached['_had_invalid_names'] );
-			} elseif ( get_option( 'properf_table_name_encoding_warning' ) ) {
-				delete_option( 'properf_table_name_encoding_warning' );
-			}
 			return $cached;
 		}
 
@@ -267,6 +265,16 @@ class ProPerf_Data_Collector {
 		}
 		$all_plugins    = get_plugins();
 		$active_plugins = get_option( 'active_plugins', array() );
+		if ( is_multisite() ) {
+			// Network-activated plugins live in a site option, not this site's
+			// active_plugins — without this they're miscounted as inactive.
+			$active_plugins = array_unique(
+				array_merge(
+					$active_plugins,
+					array_keys( get_site_option( 'active_sitewide_plugins', array() ) )
+				)
+			);
+		}
 		$active_keys    = array_intersect( $active_plugins, array_keys( $all_plugins ) );
 		$active_count   = count( $active_keys );
 		$inactive_count = count( $all_plugins ) - $active_count;
@@ -284,6 +292,28 @@ class ProPerf_Data_Collector {
 			foreach ( $hook->callbacks as $priority_callbacks ) {
 				$hook_count += count( $priority_callbacks );
 			}
+		}
+
+		$result = array(
+			'active_plugin_count'   => $active_count,
+			'inactive_plugin_count' => $inactive_count,
+			'hook_count'            => $hook_count,
+		);
+		set_transient( self::PLUGIN_METRICS_CACHE_KEY, $result, HOUR_IN_SECONDS );
+		return $result;
+	}
+
+	/**
+	 * Collect per-table DB sizes, the total DB size, and any table-name encoding warning.
+	 * Cached separately from plugin stats: this data changes on general DB writes,
+	 * not plugin (de)activation, so it isn't busted by plugin-toggle hooks.
+	 *
+	 * @return array {db_table_sizes, total_db_size_mb, table_name_warning}
+	 */
+	private function collect_table_stats() {
+		$cached = get_transient( self::SERVER_METRICS_CACHE_KEY );
+		if ( false !== $cached ) {
+			return $cached;
 		}
 
 		global $wpdb;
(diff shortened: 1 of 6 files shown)


includes/class-data-collector.php
```diff
@@ -314,6 +411,7 @@ public function collect_and_push() {
 		require_once PROPERF_DIR . 'includes/class-bigquery-client.php';
 
 		delete_transient( self::WOO_METRICS_CACHE_KEY );
+		delete_transient( self::SERVER_METRICS_CACHE_KEY );
```
pokhiii: **Real staleness bug** — the autoload metrics cache is missing from this bust list.

`collect_and_push()` deletes `WOO_METRICS_CACHE_KEY` and `SERVER_METRICS_CACHE_KEY` before calling `get_data()`, but `AUTOLOAD_METRICS_CACHE_KEY` is not deleted. `get_data()` then calls all three collectors:

- `collect_woo_order_metrics()` — cache was busted, runs live ✓
- `collect_server_metrics()` — cache was busted, runs live ✓
- `collect_autoloaded_options()` — cache still warm, returns whatever was there from the last dashboard view (up to 1 hour stale)

So the BigQuery row's `timestamp_utc` reflects the push time, but `autoloaded_option_count` and `autoloaded_option_size` can be up to an hour behind. Not catastrophic — autoload count changes slowly — but the row is internally inconsistent.

One-line fix:

```php
delete_transient( self::WOO_METRICS_CACHE_KEY );
delete_transient( self::SERVER_METRICS_CACHE_KEY );
delete_transient( self::AUTOLOAD_METRICS_CACHE_KEY );
```

Same reasoning as why WOO + SERVER are busted here in the first place: pushes should always run live queries so BigQuery history reflects real state, not what happened to be cached.

Related (worth handling in the same pass): plugin `activated_plugin` / `deactivated_plugin` hooks bust `SERVER_METRICS_CACHE_KEY` but not `AUTOLOAD_METRICS_CACHE_KEY`. Many plugins add autoloaded options on activation — an installed plugin's autoload footprint won't be reflected on the dashboard until the cache TTL expires. If you agree with the fix here, adding autoload to the activation-hook bust makes both paths symmetric.

includes/class-data-collector.php
```diff
@@ -406,12 +504,18 @@ function ( $r ) use ( $archival_ts ) {
...
 
-	/**
-	 * Delete the cached WooCommerce metrics snapshot.
-	 * Called by settings hooks when archival date or threshold changes.
-	 */
-	public static function bust_metrics_cache() {
+	public static function bust_woo_metrics_cache() {
 		delete_transient( self::WOO_METRICS_CACHE_KEY );
+		delete_transient( self::AUTOLOAD_METRICS_CACHE_KEY );
+	}
```
pokhiii, line 559: Scoping / naming concern (non-blocking).

`bust_woo_metrics_cache()` also busts the autoload metrics cache. That's what the PR description says is intentional ("WooCommerce setting changes only bust WooCommerce and autoload caches"), but two things stand out:

**1. Name doesn't match scope.** A method called `bust_woo_metrics_cache` that also invalidates a non-WooCommerce cache reads as a bug to the next developer. If the coupling is intentional, either rename (`bust_options_scope_caches()`?) or split into `bust_woo_metrics_cache()` + `bust_autoload_metrics_cache()` and call both from the hook registration site.

**2. Why does saving a WooCommerce archival date invalidate the autoload snapshot?** Autoload count changes only when options with `autoload=yes` are added/removed. Saving `properf_last_archival_date` updates one row's value — count stays constant. A recount isn't necessary. If the concern is that saving *creates* the row on first save (bumping count by 1), that's a single-integer change that self-corrects on the next natural bust. Busting the whole autoload cache for that feels aggressive.

Even if you keep the coupling, the naming split is worth doing. Both `bust_woo_metrics_cache` (WOO only) and `bust_autoload_metrics_cache` (AUTOLOAD only) as separate methods, and wire whichever combinations you actually want at the call sites.

includes/class-data-collector.php
```diff
@@ -236,6 +246,93 @@ public function collect_woo_order_metrics() {
...
+		global $wp_filter;
+		$hook_count = 0;
+		foreach ( $wp_filter as $hook ) {
+			if ( ! ( $hook instanceof WP_Hook ) ) {
+				continue;
+			}
+			foreach ( $hook->callbacks as $priority_callbacks ) {
+				$hook_count += count( $priority_callbacks );
+			}
+		}
```
pokhiii, line 295: Semantic note on the hook count metric (not a bug — worth documenting).

Issue #11 phrases this metric as *"hook count per page"*. What this loop actually measures is the count of callbacks in `$wp_filter` **at the moment `collect_server_metrics()` is called** — which depends heavily on when it runs:

- From cron: WP-Cron context. `plugins_loaded`, `init`, and admin-side hooks are registered; frontend template hooks (widgets, sidebars, template loops) are not. Some plugins register hooks lazily inside `template_redirect` or later — those won't be counted.
- From dashboard render: admin context. Different set of hooks than cron.
- From a hypothetical frontend hit: yet another set (theme hooks, plugin frontend integrations).

So `hook_count` is directional — "this site has a lot of stuff registered by the point cron fires" — but not literally "hooks per page." For a signal designed to say *"plugin stack is heavy"*, this is fine and probably even preferable (stable snapshot at a known point rather than fluctuating per URL). But it's worth a one-line comment on this method — or on the metric definition in #11 — so a future reader knows what the number means and doesn't hunt for the discrepancy.

Suggested docblock addition:

```php
 * Note: this counts callbacks registered in $wp_filter at the moment this method
 * runs (typically cron context or admin dashboard). It's not "hooks per frontend
 * page" — frontend-only hooks registered later in the request lifecycle aren't
 * counted. Meant as a directional signal for plugin-stack heaviness.
```

properf-wordpress-adapter.php
```diff
@@ -16,7 +16,7 @@
 
 define( 'PROPERF_DIR', plugin_dir_path( __FILE__ ) );
 define( 'PROPERF_URL', plugin_dir_url( __FILE__ ) );
-define( 'PROPERF_VERSION', '1.1.0' );
+define( 'PROPERF_VERSION', '1.1.1' );
```
pokhiii: SemVer nit: this PR adds a new feature area (server metrics collection, two new dashboard sections, 4 new BigQuery columns). Per SemVer this should be a minor bump (1.2.0), not a patch (1.1.1).

Patch is reserved for backwards-compatible bug fixes only. The new BigQuery columns and new collector method aren't bug fixes — they're additive features.

(The version constant on line 19 and the plugin header on line 6 both need updating.)

pokhiii (commented):
Substantial PR — closes #11 (the parent server-side-metrics tracking issue). Three new metric groups (plugin count, hook count, per-table DB sizes), a new collector, new dashboard sections, cache scoping. The 18-item "Code quality / correctness fixes" section is genuinely thorough — `array_intersect` instead of `validate_active_plugins()`, `iconv` false-return handling, `WP_Hook` instance guard, `$_POST` empty guard, encoding-warning state embedded in the cache — this is the self-review discipline paying off across multiple PRs now.

A few real findings inline. One is a staleness bug in the BigQuery push path; the others are spec drift and scoping / naming concerns.

## AC verification against #11

Issue #11 lists three metrics to collect + push:

- ✅ Active + inactive plugin count → pushed as `active_plugin_count` + `inactive_plugin_count`
- ✅ Hook count per page → pushed as `hook_count` (see inline note about "per page" semantics)
- ⚠ **All database table sizes (information_schema) → only `total_db_size_mb` is pushed.** Per-table sizes are dashboard-only. Real spec drift — #11's "Expected outcome" says *"All database table sizes (information_schema)"* under "To be added", and the whole "What you can decide after collecting it" rationale hinges on **cross-client trending of specific tables in BigQuery** to catch "unexpected growth from logging plugins, session tables, or revision accumulation." With only the total, that analysis can't happen downstream.

Two paths forward:

1. Push per-table sizes as a repeated / STRUCT field (BigQuery supports `ARRAY<STRUCT<table_name STRING, size_mb FLOAT>>`), or as a JSON-encoded string in a single column. Preserves the requirement literally.
2. Update #11 to reflect that only aggregate + WooCommerce breakdown is pushed, and per-table stays dashboard-only. Would need a note on why (row bloat? cardinality?).

Asking to pick one, not both — this is the last unchecked AC on #11 and shouldn't stay ambiguous.

## WP best-practices audit

Clean:
- All SQL through `$wpdb->prepare()` (even with `DB_NAME` which is a constant).
- Output escaping consistent across new dashboard sections.
- New option keys registered/used, no autoload bloat concerns except one small nit (see inline on the encoding-warning option).
- HPOS / WC-active handling preserved.
- Cache-busting hooks correctly moved into `register_persistent_hooks()` so WP-CLI and REST paths trigger them — this closes the concern I'd have raised about `activated_plugin` only firing on admin.

Notes:
- **Cache-busting scope**: the split into `bust_woo_metrics_cache()` and `bust_server_metrics_cache()` is the right shape, but `bust_woo_metrics_cache()` also busts the autoload cache. See inline — worth either separating or renaming.
- **`collect_and_push()` missing autoload bust**: the actual staleness bug. See inline.
- **Hook count context**: `$wp_filter` at the moment this runs isn't necessarily "per page." See inline.
- **Version bump**: 1.1.0 → 1.1.1 is a patch bump; this PR adds a new feature area, which by SemVer should be 1.2.0. Minor nit.

## Positive callouts worth naming explicitly

- **`array_intersect` avoiding `validate_active_plugins()`** — that function writes back to the DB. Using it inside a cron collector would have been a real correctness issue. Nice catch.
- **`iconv( ..., '//IGNORE' )` return values** — the `false === $clean || '' === $clean` guard is exactly right. `iconv` returning `false` on missing extension would silently corrupt array keys otherwise.
- **`instanceof WP_Hook` guard** — small change, prevents a PHP 8 fatal when legacy plugins register raw arrays into `$wp_filter`. Good defensive add.
- **Encoding warning state embedded inside the cached value** — makes the dashboard notice track the cache correctly on both hit and miss paths.
- **Duplicate `properf_bq_last_sync` write removed** from the BigQuery client — good spot; that had been a source of subtle write drift between the collector and the client.
- **The four AC items left unchecked** in the PR description body (non-WC render, cache scoping under settings changes, WP-CLI unit setting, encoding warning appearing/clearing) — I appreciate the honest labeling. Better than green-checking untested claims.

Ajay-Singh-Adhikari:
> Substantial PR — closes #11 (the parent server-side-metrics tracking issue). Three new metric groups (plugin count, hook count, per-table DB sizes), a new collector, new dashboard sections, cache scoping. The 18-item "Code quality / correctness fixes" section is genuinely thorough — `array_intersect` instead of `validate_active_plugins()`, `iconv` false-return handling, `WP_Hook` instance guard, `$_POST` empty guard, encoding-warning state embedded in the cache — this is the self-review discipline paying off across multiple PRs now.
> 
> A few real findings inline. One is a staleness bug in the BigQuery push path; the others are spec drift and scoping / naming concerns.
> 
> ## AC verification against #11
> Issue #11 lists three metrics to collect + push:
> 
> * ✅ Active + inactive plugin count → pushed as `active_plugin_count` + `inactive_plugin_count`
> * ✅ Hook count per page → pushed as `hook_count` (see inline note about "per page" semantics)
> * ⚠ **All database table sizes (information_schema) → only `total_db_size_mb` is pushed.** Per-table sizes are dashboard-only. Real spec drift — #11's "Expected outcome" says _"All database table sizes (information_schema)"_ under "To be added", and the whole "What you can decide after collecting it" rationale hinges on **cross-client trending of specific tables in BigQuery** to catch "unexpected growth from logging plugins, session tables, or revision accumulation." With only the total, that analysis can't happen downstream.
> 
> Two paths forward:
> 
> 1. Push per-table sizes as a repeated / STRUCT field (BigQuery supports `ARRAY<STRUCT<table_name STRING, size_mb FLOAT>>`), or as a JSON-encoded string in a single column. Preserves the requirement literally.
> 2. Update #11 to reflect that only aggregate + WooCommerce breakdown is pushed, and per-table stays dashboard-only. Would need a note on why (row bloat? cardinality?).
> 
> Asking to pick one, not both — this is the last unchecked AC on #11 and shouldn't stay ambiguous.
> 
> ## WP best-practices audit
> Clean:
> 
> * All SQL through `$wpdb->prepare()` (even with `DB_NAME` which is a constant).
> * Output escaping consistent across new dashboard sections.
> * New option keys registered/used, no autoload bloat concerns except one small nit (see inline on the encoding-warning option).
> * HPOS / WC-active handling preserved.
> * Cache-busting hooks correctly moved into `register_persistent_hooks()` so WP-CLI and REST paths trigger them — this closes the concern I'd have raised about `activated_plugin` only firing on admin.
> 
> Notes:
> 
> * **Cache-busting scope**: the split into `bust_woo_metrics_cache()` and `bust_server_metrics_cache()` is the right shape, but `bust_woo_metrics_cache()` also busts the autoload cache. See inline — worth either separating or renaming.
> * **`collect_and_push()` missing autoload bust**: the actual staleness bug. See inline.
> * **Hook count context**: `$wp_filter` at the moment this runs isn't necessarily "per page." See inline.
> * **Version bump**: 1.1.0 → 1.1.1 is a patch bump; this PR adds a new feature area, which by SemVer should be 1.2.0. Minor nit.
> 
> ## Positive callouts worth naming explicitly
> * **`array_intersect` avoiding `validate_active_plugins()`** — that function writes back to the DB. Using it inside a cron collector would have been a real correctness issue. Nice catch.
> * **`iconv( ..., '//IGNORE' )` return values** — the `false === $clean || '' === $clean` guard is exactly right. `iconv` returning `false` on missing extension would silently corrupt array keys otherwise.
> * **`instanceof WP_Hook` guard** — small change, prevents a PHP 8 fatal when legacy plugins register raw arrays into `$wp_filter`. Good defensive add.
> * **Encoding warning state embedded inside the cached value** — makes the dashboard notice track the cache correctly on both hit and miss paths.
> * **Duplicate `properf_bq_last_sync` write removed** from the BigQuery client — good spot; that had been a source of subtle write drift between the collector and the client.
> * **The four AC items left unchecked** in the PR description body (non-WC render, cache scoping under settings changes, WP-CLI unit setting, encoding warning appearing/clearing) — I appreciate the honest labeling. Better than green-checking untested claims.

On the per-table sizes AC drift — went with narrowing #11's scope rather than adding per-table sizes to BigQuery. Updated #11: per-table breakdown stays wp-admin dashboard-only (Top 10 Database Tables section already covers it for a single-client audit), and only `total_db_size_mb` gets pushed to BigQuery. The cross-client trending use case in #11's original write-up is real, but nothing concrete has come up needing it yet — documented in #11 that it's worth revisiting with a `REPEATED RECORD`/JSON column if that changes, rather than building it speculatively now.

Also fixed the other three:
- Autoload cache staleness in `collect_and_push()` — now busts all three transients before pushing.
- Split `bust_woo_metrics_cache()` into WooCommerce-only, with a new `bust_autoload_metrics_cache()` — wired the latter into the plugin activation/deactivation hooks alongside the server metrics bust, since installing a plugin can add autoloaded options.
- Added the docblock on `hook_count` semantics (context-dependent snapshot, not literally per-frontend-page).
- Version bumped to 1.2.0 per SemVer.

includes/class-data-collector.php
pokhiii: Verified — `collect_and_push()` at `class-data-collector.php:460` now calls `self::bust_metrics_cache()`, which composes all four bust methods (WOO + AUTOLOAD + PLUGIN + SERVER). Autoload can no longer be stale on push. Plugin lifecycle hooks (`activated_plugin`, `deactivated_plugin`, `deleted_plugin`) at `class-admin-settings.php:34-39` also bust both plugin AND autoload caches, so an installed plugin's autoloaded options show up immediately.

Bonus catch in the same commit: the activation-hook path in `properf-wordpress-adapter.php:31` also uses the aggregate `bust_metrics_cache()` — so a plugin update from any origin (WP admin, WP-CLI, unattended upgrades) shows fresh data on the first dashboard open. Resolving from my side.

includes/class-data-collector.php
pokhiii: Verified — split into four focused methods at `class-data-collector.php:553-566`:

- `bust_woo_metrics_cache()` — WOO only
- `bust_autoload_metrics_cache()` — AUTOLOAD only
- `bust_plugin_metrics_cache()` — PLUGIN only
- `bust_server_metrics_cache()` — SERVER only

`bust_metrics_cache()` composes all four. Hook registration at `class-admin-settings.php:27-39` wires each event only to the specific caches that actually change (WC settings → WOO, plugin lifecycle → PLUGIN + AUTOLOAD).

Bonus: also went further than I suggested by splitting the collector itself — `collect_server_metrics()` is now `collect_plugin_stats()` + `collect_table_stats()` with independent cache keys (`PLUGIN_METRICS_CACHE_KEY` + `SERVER_METRICS_CACHE_KEY`). Plugin toggles no longer force an `information_schema` rescan; DB writes don't force plugin recounts. Cleaner separation of invalidation triggers than I'd asked for. Resolving from my side.

includes/class-data-collector.php
pokhiii: Verified at `class-data-collector.php:282-285` — the inline comment above the `$wp_filter` loop now reads:

> Counts callbacks registered in $wp_filter at the moment this method runs (typically cron context or admin dashboard). Not "hooks per frontend page" — frontend-only hooks registered later in the request lifecycle aren't counted. Meant as a directional signal for plugin-stack heaviness.

Exactly the framing. Anyone reading this method next month will know what the number represents without hunting for the discrepancy. Resolving from my side.

properf-wordpress-adapter.php
pokhiii: Verified — `1.1.1 → 1.2.0` in both the plugin header (`properf-wordpress-adapter.php:6`) and the `PROPERF_VERSION` constant (line 19). Matches SemVer intent for an additive feature. Resolving from my side.

pokhiii (approved):
All four inline items closed — per-thread verifications posted. Two things worth calling out at the summary level:

**The per-table BQ decision is properly documented.** Instead of just skipping the AC, issue #11's body was updated with an explicit "Scope decision (2026-07-21)" block that names the tradeoff, explains *why* ("cross-client trending of specific tables isn't built here; there's no concrete use case for it yet"), and states the re-open condition ("Revisit this — and push per-table data as a `REPEATED RECORD`/JSON column — if that cross-client use case actually comes up"). That's exactly how a spec-vs-implementation deviation should be resolved — the decision lives in the source of truth, not scattered across a PR thread.

**Seven additional self-review ("R3") catches** worth naming — several are subtle bugs I would not have caught from the diff alone:

- **Multisite undercounts active plugins** — reading only per-site `active_plugins` missed network-activated plugins living in `active_sitewide_plugins`. Now merged when `is_multisite()`. Genuinely subtle — this only surfaces on multisite installations, easy to ship undetected.
- **`iconv()` unguarded on hosts without ext-iconv** — this is the sharper catch. `iconv` throws a PHP `Error`, not an `Exception`, on missing extension. The surrounding `catch ( Exception $e )` in both the dashboard render and the cron push doesn't catch `Error` — so the whole page/push would fatal instead of degrading. Now `function_exists('iconv')`-guarded the same way `get_plugins()` is.
- **`db_table_sizes` key collision** — two corrupted table names sanitizing to the same string overwrote each other while `total_db_size_mb` counted both. Total no longer summed to the visible rows. Now accumulates.
- **Fully-invalid table names skipped from the total** — a table iconv couldn't clean at all silently undercounted real DB size. Now the size is always added to the total; only the per-table display row is skipped.
- **`properf_table_name_encoding_warning` option retired** — was being autoloaded (no `false` argument), and required a get/update/delete sync dance on every cache read. Moved to the metrics snapshot itself. Nice touch given this plugin's job is measuring autoload footprint.
- **Settings field double-cached WooCommerce metrics** — `order_itemmeta_alert_threshold_mb_field()` was calling `collect_woo_order_metrics()` and setting the transient itself, redundant with the collector's own cache. Removed.
- **`deleted_plugin` and `upgrader_process_complete` hooks added** — plugin count changes on install/delete, not just activate/deactivate.

The `iconv` Error/Exception distinction and the multisite fix in particular reflect the kind of end-to-end thinking ("what happens on a host / topology I haven't personally tested?") that catches production bugs before they ship.

Approving.

